PR #1, opened 2015-08-21
Renaming and publishing

timholy:


cormullion:
Hi Tim. Does this mean that you can't say `color("red")` like you could before in Color.jl?

timholy:
To be more consistent with the directions that Julia has been going in, it's now `parse(Color, "red")`. (Julia has deprecated `int("37")`.) But in https://github.com/dcjones/Compose.jl/pull/137 I [added it to Compose](https://github.com/dcjones/Compose.jl/blob/77730f3a27d23bc15f0a2e8f700259ea3a1d796b/src/Compose.jl#L302-L304).

If folks really want this back in the general `Colors` package, it's open for discussion. I was on the fence about this one. I thought about building in a deprecation warning, but given that I've volunteered (stupidly :wink:) to fix all the packages myself, I figured I'd just break it. Finally, `sed -i -E 's/\bcolor\("/parse(Color, "/g' files...` seems to do a good job of making this transition automatically.

Anyway, do feel free to discuss.

cormullion:
It would be more user-friendly to be able to specify colors more easily than `parse(Color, "red")`, in plotting environments for example. What about `colorname("red")` or `namedcolor("red")`?

I'm only complaining because I use `color("red")` a lot... :)

timholy:
Valid complaint. Note that `color` is exported by Compose, so it will continue to work if you're plotting with Gadfly. Or are you talking about other uses?

`color` is great for interactive usage. OTOH one thing I like about this change is it helps emphasize that `color("red")` is a poor choice for when you're writing functions:
```jl
julia> foo1() = RGB(1,0,0)
foo1 (generic function with 1 method)

julia> foo2() = parse(Color, "red")
foo2 (generic function with 1 method)

julia> @time (for i = 1:10^5; foo1(); end)
  0.000001 seconds

julia> @time (for i = 1:10^5; foo2(); end)
  0.220873 seconds (2.00 M allocations: 50.354 MB, 7.53% gc time)
```
One option is we could create a `Color` string literal macro:
```jl
julia> macro Color_str(ex)
           isa(ex, AbstractString) || error("Color macro only works with literal strings")
           col = parse(Color, ex)
           :($col)
       end

julia> Color"red"
RGB{U8}(1.0,0.0,0.0)

julia> foo3() = Color"red"
foo3 (generic function with 1 method)

julia> foo3()
RGB{U8}(1.0,0.0,0.0)

julia> @time (for i = 1:10^5; foo3(); end)
  0.000001 seconds
```
How's that?

timholy:
CC @dcjones.

cormullion:
For my own purposes (I'm not a Compose-r) I can pinch your definition from Compose.jl. What about usrers of all those other plotting environments, though...

The Color string literal looks cool, actually. A bit quirky, but cool, and less typing! I could grow fond of it... :)

SimonDanisch:
I'm in strong favor for the color string macro.
It's very julian, fast and looks good!
Though, `color"red"` seems more consistent with other string macros.
This would entail color"#232323" right?

timholy:
> Though, color"red" seems more consistent with other string macros.

There's `MIME"png"`. I lean towards capitalized as being more consistent with the type `Color`, but am open to persuasion.

> This would entail color"#232323" right?

yes, that works too.

PR #2, opened 2015-08-21
replace "this library" by "Colors.jl"

ggggggggg:
I read this to mean `hsla()` has no notion of transparency at first, this phrasing is less ambiguous.


timholy:
Thanks for catching this! This sentence was outdated before I even started on refactoring, but I didn't catch it. I just deleted it in #5.

PR #4, opened 2015-08-22
Add Color string-literal macro: Color"red"

timholy:
Implements the approach discussed in https://github.com/JuliaGraphics/Colors.jl/pull/1#issuecomment-133358533

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,10 @@ The available colorspaces are described in detail in ColorTypes; briefly, the de
 
 ## Color Parsing
 
-`color(desc::String)`
+```jl
+c = Color"red"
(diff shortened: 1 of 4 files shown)


timholy:
I should have mentioned that it also adds a deprecation warning for `color("red")`.

CCing @dcjones

kmsquire:
I think I would also prefer `color"red"` to `Color"red"`. The latter somehow suggests a closer relationship to the abstract type `Color` than actually exists.

Keno:
+1 to `color"red`"

timholy:
Closing in favor of #5. Let's continue the discussion there.

PR #5, opened 2015-08-23
Second renaming, add Colorant string macro, drop Graphics dependency

timholy:
Probably the only thing to discuss is whether the literal string macro should be called `Colorant"red"` or `colorant"red"`. Since it also works for strings like `"rgba(1,0,0,0.5)"`, I'm just going to preemptively say that anyone who wishes it could be called `"Color"` or `"color"` is going to have to be extremely persuasive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,7 @@ The available colorspaces are described in detail in ColorTypes; briefly, the de
 
 `color(desc::String)`
 
-Parse a [CSS color specification](https://developer.mozilla.org/en-US/docs/CSS/color). It will
-parse any CSS color syntax with the exception of `transparent`, `rgba()`,
-`hsla()` (since this library has no notion of transparency), and `currentColor`.
+Parse a [CSS color specification](https://developer.mozilla.org/en-US/docs/CSS/color). It will parse any CSS color syntax with the exception of `currentColor`.
 
 All CSS/SVG named colors are supported, in addition to X11 named colors, when
 their definitions do not clash with SVG.
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ The available colorspaces are described in detail in ColorTypes; briefly, the de
 ## Color Parsing
 
 ```jl
-c = Colorant"red"
+c = colorant"red"
 c = parse(Colorant, "red")
 ```
 
@@ -60,7 +60,7 @@ Returns a `RGB{U8}` color, unless:
     - `"hsla(h,s,l,a)"` was used, in which case an `HSLA` color;
     - a specific `Colorant` type was specified in the first argument
 
-When writing functions the `Colorant"red"` version is preferred, because
+When writing functions the `colorant"red"` version is preferred, because
 the slow step runs when the code is parsed (i.e., during compilation
 rather than run-time).
 
--- a/README.md
+++ b/README.md
@@ -44,7 +44,10 @@ The available colorspaces are described in detail in ColorTypes; briefly, the de
 
 ## Color Parsing
 
-`color(desc::String)`
+```jl
+c = Color"red"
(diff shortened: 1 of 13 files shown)


timholy:
Several votes for lowercase over in #4. But since the string macro converts it to an object that is a subtype of `Colorant`, this seems to be the `Int(x)` vs `int(x)` thing all over again---and base julia has gone with `Int(x)`. Can folks who prefer the lowercase please provide a little more justification?

SimonDanisch:
Well, it's not a constructor in the strict sense and uppercase is reserved for constructors...
That's why I find it confusing, as I expect a call but have a string macro instead.

cormullion:
>"I'm just going to preemptively say that anyone who wishes it could be called "Color" or "color" is going to have to be extremely persuasive."

So the command to choose the color red is going to be called `colorant"red"`‽ I'm not going to try and persuade you, but that looks most peculiar to me. Still, if it's well-documented people can't complain that it's too obscure... :)

timholy:
We do have `MIME"png"`.

But in some ways I guess this is more analogous to `float(x)`, since it converts to an `AbstractFloat`, whereas `Int` is a concrete type. So I guess I can go with lowercase.

@cormullion, since it would be silly to have both `color"red"` and `colorant"red"`, it seems you have to go with the more general name. For example, suppose we had a string macro for real numbers; calling it `int"3.27"` rather than `real"3.27"` would be pretty nonsensical, even though you might often wish for `int"15"`.

timholy:
Incidentally, the test failure on nightly is new as of a couple of days ago---it didn't happen on my checkout of master. I'm trying to track it down.

cormullion:
>silly to have both color"red" and colorant"red"

Perhaps, although redundancy at the user-interface level is considered more acceptable than elsewhere. Hence "aliases"...

timholy:
You can certainly make an alias. But here's another way to explain it: given a `Color` `c`, `color(c)` just returns `c`; given a `TransparentColor` `ct`, `color(ct)` strips the transparency and returns a `Color`. Hence, logically `color"rgba(1.0,0,0,0.5)"` needs to strip the transparency. This is surely not what you intended.

timholy:
Very curiously, the breakage is https://github.com/JuliaLang/julia/commit/56360a7ee1ebc6028ce16f352749843bee48cf76. I'm going to be on the road for a big chunk of the day, so it will take a little more digging.

kmsquire:
> But since the string macro converts it to an object that is a subtype of `Colorant`, this seems to be the `Int(x)` vs `int(x)` thing all over again---and base julia has gone with `Int(x)`. Can folks who prefer the lowercase please provide a little more justification?

1. `colorant"red"` isn't a type constructor, so conventions that apply to type constructors don't really apply.  It does, of course, return an object which has a type, but so do many other `_str` macros, and only a few of those are capitalized.  `MIME`, for example, is an acronym (although I've seen it written in lower case as well).  Are there other capitalized `_str` macros?

2. `Colorant"red"` certainly suggests that it produces something of type `Colorant`.  But the difference in usage between `Colorant_str` and `Colorant` is subtle, especially for someone just coming to the language.  Why can't I just write `Colorant("red")`?  Alternatively, why doesn't `Integer"234"` work?  Since `Colorant"red"` is a distinctly different operation than `Colorant(a)` (ignoring that this call doesn't exist), why not make it obvious that it's a different operation?

timholy:
Yeah, I'm persuaded on the lowercase thing; in addition to your and @SimonDanisch's arguments, the `float` example is pretty compelling to me.

I'll change it, right after I figure out what's up with the test failure.

timholy:
See https://github.com/JuliaLang/julia/issues/12771.

PR #6, opened 2015-08-24
Add migration tips

timholy:
This may help people who didn't notice the post to julia-users

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Colors on julia-release](http://pkg.julialang.org/badges/Colors_release.svg)](http://pkg.julialang.org/?pkg=Colors&ver=release)
 [![Colors on julia-nightly](http://pkg.julialang.org/badges/Colors_nightly.svg)](http://pkg.julialang.org/?pkg=Colors&ver=nightly)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -302,3 +305,42 @@ Functions in this library were mostly implemented according to:
...
+
+```sh
+# Intended to be run from the top directory in a package
+# Do not run this twice on the same source tree without discarding
+# the first set of changes.
+sed -i 's/\bColor\b/Colors/g' REQUIRE
+
+fls=$(find . -name "*.jl")
+sed -i 's/\bColor\b/Colors/g' $fls               # Color -> Colors
+sed -i -E 's/\bcolor\("(.*?)"\)/colorant\"\1\"/g' $fls   # color("red") -> colorant"red"
```
waldyrious: I thought `sed` used `-r` rather than -E (as grep does). That's [what the manual says](http://www.gnu.org/software/sed/manual/sed.txt), at least. Is it a common alias in popular systems? Or an OS-based distinction perhaps?

timholy: Didn't know that, I'm a `sed` newbie. I guess I should change it.

waldyrious: So am I :) I came across this apparent error just a few days ago, here: https://github.com/tldr-pages/tldr/pull/293

timholy: Thanks. Fixed now.

Issue #15, opened 2015-09-16
colormap breaking on 0.3


rohitvarkey:
Hi,

I noticed this when my Travis build for ThreeJS.jl failed. Reproduced locally by using Julia 0.3.11 and Colors `v0.5.3`. I tried `Pkg.checkout("Colors")` also to no avail. 
```julia
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "help()" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.3.11 (2015-07-27 06:18 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-apple-darwin13.4.0

julia> using Colors

julia> colormap("RdBu")
ERROR: type cannot be constructed
 in mix_linearly at /Users/rohitvarkey/.julia/v0.3/Colors/src/colormaps.jl:132
 in sequential_palette at /Users/rohitvarkey/.julia/v0.3/Colors/src/colormaps.jl:152
 in diverging_palette at /Users/rohitvarkey/.julia/v0.3/Colors/src/colormaps.jl:225
 in colormap at /Users/rohitvarkey/.julia/v0.3/Colors/src/colormaps.jl:282
 in colormap at /Users/rohitvarkey/.julia/v0.3/Colors/src/colormaps.jl:256
```

rohitvarkey:
Works fine on `v0.5.1` though!

stevengj:
Maybe `base_color_type(C)(...)` is relying on the implicit `call` to `convert(base_color_type(C), ...)` in 0.4+?   Although the problem is probably in ColorTypes.jl, not Colors.jl, though.

stevengj:
No, I don't think it is that.  It is almost looking like a bug in 0.3.  I added some `println` statements to see what `colormap` is doing.  The error occurs in the `mix_linearly` function in `colormaps.jl`, which I changed to:
```jl
    function mix_linearly{C<:Color}(a::C, b::C, s)
        println("color $C($a,$b,$c)")
        B = base_color_type(C)
        (arg1,arg2,arg3) = ((1-s)*comp1(a)+s*comp1(b), (1-s)*comp2(a)+s*comp2(b), (1-s)*comp3(a)+s*comp3(b))
[...]

PR #19, opened 2015-09-19
Union() -> Union{}

yuyichao:


timholy:
The error seems to be #15.

PR #20, opened 2015-09-28
fixed for 0.4/Colors.jl

cormullion:
although the output SVG is a bit different (colors are in a slightly
different order from before), this particular file is only a
quick-reference/pretty handout, so the changes aren’t significant… :)


cormullion:
failure is due to #15 I think

timholy:
Out of curiosity, what error did this fix?

cormullion:
@timholy Well for one thing I left `#!/Applications/Julia-0.3.9.app...` at the top of the file. (Bad habit I have... :)) Also I changed the color comparison code (`hslx = convert(LUV, RGB(rx, gx, bx))` 😱). But the failure of the tests is due to the colormap issues in #15.

timholy:
I guess what I meant is: what is the motivation for this PR? Was it a deprecation warning, an error message, an incorrect answer, or something else?

cormullion:
Your revision of Aug 22 changed Color to Colors, but then the file failed with the error:

    ERROR: LoadError: UndefVarError: LUV not defined
     in compare_colors 

(after changing the shebang line). I thought I'd better fix it (and improve it while I was there).

timholy:
OK, thanks!

PR #22, opened 2015-10-12
UFixed transition and commit to julia 0.4

timholy:
Incorporate and fix deprecations caused by the wonderful JeffBezanson/FixedPointNumbers.jl#27. This also commits to julia 0.4 and deletes cruft.

This won't pass tests until JeffBezanson/FixedPointNumbers.jl#31 gets tagged. I'll re-run the travis jobs when that happens, but submitting for comments.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,4 @@
-julia 0.3
-ColorTypes
-FixedPointNumbers
-Compat 0.2
+julia 0.4
+ColorTypes 0.2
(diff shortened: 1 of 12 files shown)


vchuravy:
@timholy I workinf on merging UFixed and Fixed right now since effectively the code is duplicated now. I should have something ready tomorrow or so.

timholy:
OK, I'll wait for these changes. Might as well get all the API breakage out of the way at once.

Issue #232, opened 2015-11-16
Graceful parsing of Colorants


tbreloff:
This isn't ideal behavior:

```
julia> using Colors

julia> parse(Colorant, "red")
RGB{U8}(1.0,0.0,0.0)

julia> parse(Colorant, colorant"red")
ERROR: StackOverflowError:
 in parse at /home/tom/.julia/v0.4/Colors/src/parse.jl:136 (repeats 342 times)
```

Should there be a better signature in [parse](https://github.com/JuliaGraphics/Colors.jl/blob/master/src/parse.jl#L136):

```
Base.parse{C<:Colorant}(::Type{C}, desc::AbstractString) = convert(C, parse(Colorant, desc))::C
```

and/or should there be a "pass-through" option for Colorants:

```
Base.parse{C<:Colorant}(::Type{C}, c::Colorant) = convert(C, c)
```

timholy:
FYI `colorant"red"` by itself (without the parse) is all you need. But a pass-through would be great, if you want to submit that.

tbreloff:
Yes... The issue came up somewhere in Plots, of course I'm still looking for the bug since the full stack trace isn't printed due to the stack overflow. (Is that a bug in julia?)

I'll put together a PR for the signatures I mentioned.

timholy:
Ugh, StackOverflows shouldn't happen even in cases of user error, but I often forget to watch out for this.

The key technique is probably to change only one argument type at a time. Comparing https://github.com/JuliaGraphics/Colors.jl/blob/830936435de28ec1989728cbbde36f335d760a73/src/parse.jl#L75 and https://github.com/JuliaGraphics/Colors.jl/blob/830936435de28ec1989728cbbde36f335d760a73/src/parse.jl#L136, I'd predict the StackOverflow should go away if you add `::AbstractString` to the declaration of `desc` in the latter. If this is true, would be good to change this.

timholy:
Fixed by #233

PR #233, opened 2015-11-16
fixed parse StackOverflow and added Symbol/Colorant versions; closes …

tbreloff:
…#232

Let me know if you'd like any adjustments or additional tests.

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -72,7 +72,7 @@ Returns:
     - "hsla(h,s,l,a)" was used, in which case an `HSLA` color;
     - a specific `Colorant` type was specified in the first argument
 """
-function Base.parse(::Type{Colorant}, desc::AbstractString)
+function _parse_colorant(desc::AbstractString)
     desc_ = replace(desc, " ", "")
     mat = match(col_pat_hex2, desc_)
     if mat != nothing
@@ -133,7 +133,17 @@ function Base.parse(::Type{Colorant}, desc::AbstractString)
     return RGB{U8}(c[1] / 255, c[2] / 255, c[3] / 255)
 end
 
-Base.parse{C<:Colorant}(::Type{C}, desc) = convert(C, parse(Colorant, desc))::C
+function Base.parse{C<:Colorant}(::Type{C}, desc::AbstractString)
+    c = _parse_colorant(desc)
+    if C === Colorant
+        c
+    else
+        convert(C, c)::C
+    end
+end
+
+Base.parse{C<:Colorant}(::Type{C}, desc::Symbol) = parse(C, string(desc))
+Base.parse{C<:Colorant}(::Type{C}, c::Colorant) = C === Colorant ? c : convert(C, c)
 
 macro colorant_str(ex)
     isa(ex, AbstractString) || error("colorant requires literal strings")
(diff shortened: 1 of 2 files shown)


src/parse.jl
```diff
@@ -72,7 +72,7 @@ Returns:
     - "hsla(h,s,l,a)" was used, in which case an `HSLA` color;
     - a specific `Colorant` type was specified in the first argument
 """
-function Base.parse(::Type{Colorant}, desc::AbstractString)
+function _parse_colorant(desc::AbstractString)
```
timholy: This attaches the docstring to `_parse_colorant` when it should be attached to `Base.parse`.

src/parse.jl
```diff
@@ -133,7 +133,17 @@ function Base.parse(::Type{Colorant}, desc::AbstractString)
...
 end
 
-Base.parse{C<:Colorant}(::Type{C}, desc) = convert(C, parse(Colorant, desc))::C
+function Base.parse{C<:Colorant}(::Type{C}, desc::AbstractString)
+    c = _parse_colorant(desc)
+    if C === Colorant
+        c
+    else
+        convert(C, c)::C
+    end
```
timholy: This seems unlikely to be type-inferrable. The previous function definitions were, I think. Any particular reason you rewrote it this way?

src/parse.jl
```diff
@@ -133,7 +133,17 @@ function Base.parse(::Type{Colorant}, desc::AbstractString)
...
+end
+
+Base.parse{C<:Colorant}(::Type{C}, desc::Symbol) = parse(C, string(desc))
+Base.parse{C<:Colorant}(::Type{C}, c::Colorant) = C === Colorant ? c : convert(C, c)
```
timholy: This is also likely to be type-unstable.

src/parse.jl
tbreloff: I re-wrote this way because it didn't seem like it would work otherwise.  Hopefully this example sufficiently describes the problem:
```
julia> f(::Type{Integer}) = "first def"
f (generic function with 1 method)

julia> f{I<:Integer}(::Type{I}) = "second def"
f (generic function with 2 methods)

julia> f(Integer)  # I want this to use the first def
"second def"

julia> f(Int64)
"second def"
```

Essentially, the main function could not be called as the `Base.parse{C<:Colorant}(::Type{C}, desc::AbstractString)` method would mask the first.  If you have a good solution to the toy example I provided, then we should certainly apply that.  All your other comments stem from this same issue.

timholy: Argh, you are right, and sorry for the confusion. This is Julia bug https://github.com/JuliaLang/julia/issues/12580 or similar. In similar situations in the past I've been able to make it work by changing the order I defined the functions in (which shouldn't be important), but in this case one seems to consistently always call "second def."

Here's a workaround:
```jl
julia> f{I<:Integer}(::Type{I}) = _f(I, super(I))
f (generic function with 1 method)

julia> _f(::Type{Integer}, ::Type{Real}) = "first def"
_f (generic function with 1 method)

julia> _f{I<:Integer,I2<:Integer}(::Type{I}, ::Type{I2}) = "second def"
_f (generic function with 2 methods)

julia> f(Integer)
"first def"

julia> f(Int)
"second def"
```

This ugliness should not be necessary, but there you have it.

timholy:
Pushing additional commits to a PR doesn't send an email, so if you address review comments don't hesitate to also ping me.

Let's wait a bit more for comments on #239.

PR #239, opened 2015-12-01
Add a fallback for `parse(Colorant, ::Colorant)`

timholy:
I have some reservations about this because of the following behavior:
```jl
julia> parse(Int, "5")
5

julia> parse(Int, 5)
ERROR: MethodError: `parse` has no method matching parse(::Type{Int64}, ::Int64)
Closest candidates are:
  parse{T<:Integer}(::Type{T<:Integer}, ::Char)
  parse{T<:Integer}(::Type{T<:Integer}, ::Char, ::Integer)
  parse{T<:Integer}(::Type{T<:Integer}, ::AbstractString, ::Integer)
```
But this is one way to fix https://github.com/dcjones/Gadfly.jl/issues/756. I've got another PR I'm preparing for Gadfly that is an alternative fix. Thoughts?

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -135,6 +135,8 @@ end
 
 Base.parse{C<:Colorant}(::Type{C}, desc) = convert(C, parse(Colorant, desc))::C
 
+Base.parse{C<:Colorant}(::Type{C}, c::Colorant) = convert(C, c)::C
+
 macro colorant_str(ex)
     isa(ex, AbstractString) || error("colorant requires literal strings")
     col = parse(Colorant, ex)


timholy:
To be explicit: is the lack of such a method for `Int` deliberate, or an oversight?

tbreloff:
Tim: this is the same as #233. I don't care if that gets closed, but I think just adding the extra parse method for Colorant causes trouble somewhere... Possibly it wasn't getting called because of the dispatch bug with parametric subtypes? Anyways the PRs are redundant so you can close one of them.

timholy:
Amazing how quickly I forget.

PR #241, opened 2015-12-23
distinguishable_colors: restrict candidates to RGB-realizable colors

timholy:
I can't remember where this originally came up, but I believe that @m-lohmann or @tbreloff discovered that distinguishability improves if you ensure that all candidate colors are within the gamut of RGB.


m-lohmann:
Yes, I found out that `distinguishable_colors` produced colors outside the sRGB range (using the wrong color space) which led to colors that looked too similiar.
Thanks for fixing it!

timholy:
Thanks, @m-lohmann!

PR #242, opened 2016-02-28
Convert to Gray using round, not trunc

timholy:
Should fix https://github.com/timholy/Images.jl/pull/445#issuecomment-189866806. CC @Evizero.


Evizero:
```
INFO: Images tests passed
```

:tada:

timholy:
That's great! Thanks again.

timholy:
Must be spring already, green checkmarks are sprouting all over...

PR #249, opened 2016-07-07
Fix RGB->HSL conversion InexactError. Fixes ColorTypes #40

timholy:
CC @gabrielgellner. All I did was make sure I convert to the output eltype (e.g., `Float32`) before performing any arithmetic operations.


gabrielgellner:
Makes sense. Sorry I missed what to change! Thanks so much for fixing this.

timholy:
Not a problem, thanks for reporting with a clear test case---you're making the package better already :wink:.

timholy:
The failures on OSX are https://github.com/JuliaLang/Homebrew.jl/issues/122.

PR #253, opened 2016-09-22
generalize linspace

hofmannmartin:
I just realized that the function `linspace` can not be used to linear interpolate subtypes of `TransparentColor`.

I propose to loosen the restrictions on `T` to allow such a usage.


stevengj:
Sounds good, can you add a test?

hofmannmartin:
I just figured that the generalization makes sense only if the function `weighted_color_mean` works for all colors. The current implementation of `weighted_color_mean` just handles a hand full of cases.

If one implements the following function within the `ColorVectorSpace` package however and moves my version of `linspace` as well
```
function weighted_color_mean{T<:Real,C<:Colorant}(w1::T, c1::C, c2::C)
        w2 = w1 >= 0 && w1 <= 1 ? 1 - w1 : throw(DomainError())
        return w1*c1+w2*c2
end
```
everything should work out fine. What do you think?

timholy:
Supporting transparency is very reasonable. `weighted_color_mean` (an "old" function) could be written more generally like this:
```jl
# We have to make sure the two colorants are in the same colorspace. This version doesn't require that the two
# also have the same numeric type (here, one could be UFixed8 and other Float64)
function weighted_color_mean(w1::Real, c1::Colorant, c2::Colorant)
    w2 = w1 >= 0 & w1 <= 1 ? 1 - w1 : throw(DomainError())
    _weighted_color_mean(base_colorant_type(c1), base_colorant_type(c2), w1, w2, c1, c2)

function _weighted_color_mean{C<:Color3}(::Type{C}, ::Type{C}, w1, w2, c1, c2)
    C(w1*comp1(c1)+c2*comp1(c2), w1*comp2(c1)+w2*comp2(c2), w1*comp3(c1)+w2*comp3(c2))
end

_weighted_color_mean(::Type, ::Type, w1, w2, c1, c2) =
    throw(ArgumentError("the two colors must be from the same colorspace, but got $c1 and $c2"))
```
with additional versions for 1, 2, and 4-component colors (`AbstractGray`, `TransparentGray`, and `Color4`).

Going the ColorVectorSpace route will be less comprehensive, because it only supports grayscale and RGB (plus their transparent versions). This is a deliberate decision, because addition of colors is conceptually fraught: in general `c1+c2` does not give the same result as if you convert both to a different colorspace, perform the addition, and then convert back to the original colorspace. In other words., with two exceptions (XYZ and LMS) colorspaces are nonlinear. ColorVectorSpace exists to "force" that linearity to RGB, basically saying "I don't care that it's not really correct." The package exists because RGB "math" is so widely used for image processing, and performance would stink if you did intermediate conversions to a linear colorspace. But that's why it's a separate package, making an "optional add on" that the purists (i.e., people who really know their colorimetry) can avoid.

`weighted_color_mean` is much more specialized and designed specifically to "not care," so it's reasonable for it to support all colorspaces.

hofmannmartin:
@timholy the code you proposed works good. 
```julia
function weighted_color_mean(w1::Real, c1::Colorant, c2::Colorant)
    w2 = w1 >= 0 && w1 <= 1 ? 1 - w1 : throw(DomainError())
    _weighted_color_mean(base_colorant_type(c1), base_colorant_type(c2), w1, w2, c1, c2)
end

function _weighted_color_mean{C<:ColorTypes.AbstractGray}(::Type{C}, ::Type{C}, w1, w2, c1, c2)
    C(w1*comp1(c1)+w2*comp1(c2))
end

function _weighted_color_mean{C<:ColorTypes.TransparentGray}(::Type{C}, ::Type{C}, w1, w2, c1, c2)
    C(w1*comp1(c1)+w2*comp1(c2), w1*alpha(c1)+w2*alpha(c2))
end

function _weighted_color_mean{C<:ColorTypes.Color3}(::Type{C}, ::Type{C}, w1, w2, c1, c2)
    C(w1*comp1(c1)+w2*comp1(c2), w1*comp2(c1)+w2*comp2(c2), w1*comp3(c1)+w2*comp3(c2))
end

function _weighted_color_mean{C<:ColorTypes.Transparent3}(::Type{C}, ::Type{C}, w1, w2, c1, c2)
    C(w1*comp1(c1)+w2*comp1(c2), w1*comp2(c1)+w2*comp2(c2), w1*comp3(c1)+w2*comp3(c2), w1*alpha(c1)+w2*alpha(c2))
end

function _weighted_color_mean{C<:ColorTypes.TransparentRGB}(::Type{C}, ::Type{C}, w1, w2, c1, c2)
    C(w1*comp1(c1)+w2*comp1(c2), w1*comp2(c1)+w2*comp2(c2), w1*comp3(c1)+w2*comp3(c2), w1*alpha(c1)+w2*alpha(c2))
end

function _weighted_color_mean(::Type, ::Type, w1, w2, c1, c2) 
    throw(ArgumentError("the two colors must be from the same colorspace, but got $c1 and $c2"))
end
```

The only downside is, that the result usually has a different type.
```julia
julia> weighted_color_mean(0.5,Gray(0),Gray(1))
Gray{Float64}(0.5)
```
whereas the type of the input colors is `Gray{U8}`. Does anyone know how to fix this?

timholy:
If you specify the weight as 0.5 (a Float64), I don't think it's unreasonable to return a `RGB{Float64}`. If you specified the weight as `U8(0.5)`, and `c1` and `c2` both use `U8` element type, then it does indeed seem better to return a `RGB{U8}`. I think the key thing is the computation of `w2`, it would be better to say `oftype(w1, 1-w1)`.

timholy:
The failures are "just" code coverage. Here's how to understand what it means:

- The "codecov/project" is for the project as a whole: 65.3% of all executable lines have tests, and your changes reduce that amount by 0.53% Since this is not the direction we want to be going in, that's considered a test failure. (I think the "color suite" was around 16% when we first started measuring coverage. Through the efforts of many of us, coverage has climbed dramatically over the years, for example with ColorTypes having tests for over 90% of its lines.)
- Even more serious is "codecov/patch", which refers specifically to the lines you changed: it's telling you that 0% of those changes have an accompanying test, and so the failure highlights the fact that we have no way of telling whether this code works or is completely broken. This simply suggests that `weighted_color_mean` has never been tested (which a quick grep confirms), which is not your fault, but on the other hand has now become your problem (sorry). Where it says "target 65.82%," what that means is that it's testing whether you've at least hit the standard of the rest of the package.

In some ways writing tests is a pain, but it's totally worth it. Base has improved dramatically due to @kshyatt's testing efforts. I spent most of my weekend writing tests for (and discovering bugs in :frowning:, then fixing) insufficiently-tested functionality all over the color and image landscape of julia (e.g., https://github.com/JuliaMath/FixedPointNumbers.jl/pull/54/commits, https://github.com/JuliaMath/FixedPointNumbers.jl/pull/53/commits, https://github.com/JuliaImages/ImageFiltering.jl/pull/10/commits). This was hours and hours of work but it should translate into vastly improved experience for users.

As far as getting all color types, `subtypes(Colorant)` should do the job. More specifically you can say `subtypes(T)` where `T` is some abstract type. The `AbstractGray` and `Color3` typealiases might be handy here, and possibly unexported constants in ColorTypes like `parametric3`.

src/utilities.jl
```diff
@@ -32,34 +32,49 @@ end
...
 
-Returns a color that is the weighted mean of `c1` and `c2`, where `c1`
-has a weight 0 ≤ `w1` ≤ 1.
+Returns the color `w1*c1 + (1-w1)*c2` that is the weighted mean of `c1` and 
+`c2`, where `c1` has a weight 0 ≤ `w1` ≤ 1.
 """
-weighted_color_mean(w1::Real, c1::RGB24, c2::RGB24) =
-    convert(RGB24, weighted_color_mean(w1, convert(RGB, c1), convert(RGB, c2)))
+function weighted_color_mean(w1::Real, c1::Colorant, c2::Colorant)
+    w1 = convert(eltype(c1),w1)
```
timholy: How about `promote(eltype(c1), eltype(c2))`? It seems better to make it symmetric.

timholy: Also, give the converted value a a new variable name, otherwise this is not type-stable (if the user supplies a `Float64` and it gets converted to a `U8`, then `w1` will not have concrete type and the resulting code will be poorly optimized).

src/utilities.jl
```diff
@@ -32,34 +32,49 @@ end
...
+    w2 = w1 >= 0 && w1 <= 1 ? oftype(w1,1-w1) : throw(DomainError())
+    _weighted_color_mean(base_colorant_type(c1), base_colorant_type(c2), w1, w2, c1, c2)
+end
+
+function _weighted_color_mean{C<:ColorTypes.AbstractGray}(::Type{C}, ::Type{C}, w1, w2, c1, c2)
```
timholy: I think you can dramatically simplify all these with the new `mapc` function, `mapc((x,y)->w1*x+w2*y, c1, c2)`. (Untested but should work.) This should cover all 4 cases with a single function call.

timholy:
Good catch on using `promote_type` rather than `promote` (you read my mind).

The code is looking good, just needs a test. (And you'll catch your typo once you add the test and try running it :wink:.)

timholy:
Stupid typo on my part, sorry about that. https://github.com/JuliaGraphics/ColorTypes.jl/pull/54

timholy:
We'll actually have to tag a new version of ColorTypes for your tests to pass, so there will be a little delay.

hofmannmartin:
I did manage to write the tests.

However at the moment I do not like that I have to handle the color type `Gray{Bool}` differently. It just does not really fit in. It might be better to throw an Error if someone tries to weight two Gray{Bool}`s. Opinions appreciated.

timholy:
I think it would be fine to throw an error, but I also think this approach is fine.

Once you're happy, I'm happy to merge this. (I will probably squash it down to a single commit, if that's OK. I can do that with a single click, so it's not hard.)

hofmannmartin:
Decided to catch Gray{Bool} with a DomainError. Makes tests and code much cleaner.

I am happy now. Would be a great if you could squash everything down to a single commit. Thank you for your help. Very much appreciated.

timholy:
Thanks for tackling this! It's nice improvement.

Issue #256, opened 2016-09-27
unexpected results inconvert RGB -> [DIN99d,DIN99dA,ADIN99d] -> RGB


hofmannmartin:
Normally, I would expect that the conversion `c1::TypeA` -> `c2::TypeB` -> `c3::TypeA` yields that `c1` is approximately equal to `c3`. However for the types above one gets

``` julia
julia> convert(RGB,convert(DIN99d,RGB(1,0.5,0)))
RGB{Float64}(0.5618900556828353,0.660676085072947,0.6265900923320097)

julia> convert(RGBA,convert(DIN99dA,RGBA(1,0.5,0,0.4)))
RGBA{Float64}(0.5618900556828353,0.660676085072947,0.6265900923320097,0.4)
```

which suggests, that there is a problem with the conversion `RGB`<->`DIN99d`.

m-lohmann:
I implemented DIN99 and DIN99o back then... I’ll have a look at the DIN99d conversion.

hofmannmartin:
thank you very much

m-lohmann:
I thought I already had. Something must have gone wrong. I'll make a pull request as soon as I can.

PR #259, opened 2016-09-28
Fix DIN99d->XYZ conversion #256

m-lohmann:
rotation terms, constants, sign fixed

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -353,7 +353,7 @@ function cnvt{T}(::Type{XYZ{T}}, c::DIN99d)
 
     # Go back to C-h space
     # FIXME: Clean this up (why is there no atan2d?)
-    h = rad2deg(atan2(c.b,c.a)) + 50
+    h = rad2deg(atan2(c.b,c.a)) - 50
     while h > 360; h -= 360; end
     while h < 0;   h += 360; end
 
@@ -361,14 +361,14 @@ function cnvt{T}(::Type{XYZ{T}}, c::DIN99d)
 
     # Intermediate terms
     G = (exp(C/22.5)-1)/0.06
-    f = G*sind(h - 50)
-    ee = G*cosd(h - 50)
+    f = G*sind(h)
+    ee = G*cosd(h)
 
-    l = (exp(c.l/325.22)-1)/0.0036
+    l = (exp(c.l/325.221)-1)/0.0036
     # a = ee*cosd(50) - f/1.14*sind(50)
-    a = ee*0.6427876096865393 - f/1.14*0.766044443118978
+    a = ee*0.6427876096865394 - f/1.14*0.766044443118978
     # b = ee*sind(50) - f/1.14*cosd(50)
-    b = ee*0.766044443118978 - f/1.14*0.6427876096865393
+    b = ee*0.766044443118978 + f/1.14*0.6427876096865394
 
     adj = convert(XYZ, Lab(l, a, b))
 
@@ -643,13 +643,13 @@ function cnvt{T}(::Type{DIN99d{T}}, c::XYZ{T})
 
     # Apply L*a*b*-space correction
     lab = convert(Lab, adj_c)
-    adj_L = 325.22*log(1+0.0036*lab.l)
+    adj_L = 325.221*log(1+0.0036*lab.l)
 
     # Calculate intermediate parameters
     # ee = lab.a*cosd(50) + lab.b*sind(50)
-    ee = lab.a*0.6427876096865393 + lab.b*0.766044443118978
+    ee = lab.a*0.6427876096865394 + lab.b*0.766044443118978
     # f = 1.14*(lab.b*cosd(50) - lab.a*sind(50))
-    f = 1.14*(lab.b*0.6427876096865393 - lab.a*0.766044443118978)
+    f = 1.14*(lab.b*0.6427876096865394 - lab.a*0.766044443118978)
     G = sqrt(ee^2+f^2)
 
     # Calculate hue/chroma


src/conversions.jl
```diff
@@ -353,22 +353,22 @@ function cnvt{T}(::Type{XYZ{T}}, c::DIN99d)
...
-    f = G*sind(h - 50)
-    ee = G*cosd(h - 50)
+    f = G*sind(h)
+    ee = G*cosd(h)
 
-    l = (exp(c.l/325.22)-1)/0.0036
+    l = (exp(c.l/325.221)-1)/0.0036
     # a = ee*cosd(50) - f/1.14*sind(50)
-    a = ee*0.6427876096865393 - f/1.14*0.766044443118978
+    a = ee*0.6427876096865394 - f/1.14*0.766044443118978
```
timholy: I'll bet this change isn't a big contributor to fixing the bug :smile:

src/conversions.jl
```diff
@@ -643,13 +643,13 @@ function cnvt{T}(::Type{DIN99d{T}}, c::XYZ{T})
...
-    adj_L = 325.22*log(1+0.0036*lab.l)
+    adj_L = 325.221*log(1+0.0036*lab.l)
 
     # Calculate intermediate parameters
     # ee = lab.a*cosd(50) + lab.b*sind(50)
-    ee = lab.a*0.6427876096865393 + lab.b*0.766044443118978
+    ee = lab.a*0.6427876096865394 + lab.b*0.766044443118978
     # f = 1.14*(lab.b*cosd(50) - lab.a*sind(50))
-    f = 1.14*(lab.b*0.6427876096865393 - lab.a*0.766044443118978)
+    f = 1.14*(lab.b*0.6427876096865394 - lab.a*0.766044443118978)
```
timholy: Since we were previously passing tests despite the breakage, this fix really needs to be accompanied by tests that Colors would have formerly failed. #256 might be a good starting point?

src/conversions.jl
m-lohmann: Haha, true. I just noticed the difference while I was searching for the bug and thought it doesn’t hurt either ;)

src/conversions.jl
hofmannmartin: The Idea could be to chose a color space  which is able to represent all colors (`XYZ`?) and and then convert each color to `XYZ` and back again. This should only introduce a small error.
```julia
allowederr = 1e-2 # 1% deviation allowed
for C1 in ColorTypes.parametric3
    c1,c2,c3 = rand(3)
    color1 = C1(c1,c2,c3)
    color2 = convert(C1, convert(XYZ,color1))
    @test_approx_eq_eps(comp1(color1),comp1(color2),allowederr)
    @test_approx_eq_eps(comp2(color1),comp2(color2),allowederr)
    @test_approx_eq_eps(comp3(color1),comp3(color2),allowederr)
end
```
I am no expert on color theory though but from the mathematical point of view this is a reasonable test.

m-lohmann: I would run more than just 3 random colors just to make sure the whole color space is covered.

Some errors might only be significant in certain quadrants. The ```X’``` modification in ```DIN99d``` is not homogeneous throughout the colorspace, for example. The logarithmic Chroma compression of the ```DIN``` color spaces might cause larger or smaller errors in different regions as well. Testing e.g. 50 random samples should be a rather safe way to get everything covered.

And finally we should implement a fixed set with one-directional absolute checks as well. If the same error is in the forwards and backwards conversion the proposed random test can’t catch them.

Don’t forget that 1% in ```XYZ``` or ```Yxy``` color space can already be above ```1 ΔE``` threshold value or in the order of the size of the MacAdam color discrimination ellipses. And 1% in RGB is already a large color difference that wouldn’t be tolerable. The problem is that e.g. RGB is not perceptually uniform.

I would say that random tests with a 1% threshold are fine for ```Lab/DIN``` → ```XYZ``` → ```Lab/DIN``` checks. I would lower the threshold for ```RGB```, ```Yxy``` etc. This should be no problem at all because numerically the errors should be a lot lower.

hofmannmartin: > I would run more than just 3 random colors just to make sure the whole color space is covered.

Certainly, one needs to run multiple iterations to cover the whole color space.

> And finally we should implement a fixed set with one-directional absolute checks as well. If the same error is in the forwards and backwards conversion the proposed random test can’t catch them.

True. However to compensate you will have to implement the inverse of your error. Anyhow, it would be a good idea to include some one directional test to `XYZ` using known points and test the conversion result.

> Don’t forget that 1% in XYZ or Yxy color space can already be above 1 ΔE threshold value or in the order of the size of the MacAdam color discrimination ellipses. And 1% in RGB is already a large color difference that wouldn’t be tolerable. The problem is that e.g. RGB is not perceptually uniform.

As I pointed the exact thresholds would need to be tweaked by some color experts ;)

m-lohmann: It’s simple, if you e.g. use wrong constants in both functions. The constants in DIN conversions to and from XYZ or Lab aren’t inverses but use the same numerical values. Copy/paste the same erroneous value into both functions and you won’t notice the error.

timholy: I'm no color-theory expert, but I think the hard part about random is that the gamut of some colorspaces may not fully cover the other---some transformations will get "clipped" by the boundary of the colorspace and give large errors. So I think this might benefit from someone who knows the color theory.

I was sufficiently worried about this that I used some independent tools to generate the pairs of conversions that you find in the `test_conversions.jld` file:
```
ColorTypes.YCbCr{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.HSI{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.HSL{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.LCHab{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.LCHab{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.RGB{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.Luv{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.YCbCr{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.HSV{Float64} -> ColorTypes.HSV{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.YCbCr{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.LMS{Float64} -> ColorTypes.Lab{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.Lab{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.LMS{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.XYZ{Float64} -> ColorTypes.HSL{Float64}
ColorTypes.RGB{Float64} -> ColorTypes.HSI{Float64}
ColorTypes.Luv{Float64} -> ColorTypes.XYZ{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.YIQ{Float64}
ColorTypes.YIQ{Float64} -> ColorTypes.RGB{Float64}
```
These were all the types supported by my external tool(s).

Note that DIN is not in there. Consequently, I ask: @m-lohmann, can you please write the tests for DIN? There's no one who understands DIN better than you...

m-lohmann: The gamut clipping problem only occurs in conversions to RGB, as far as I am aware of it, because somehow someone thought it was mandatory to clip colors to the sRGB gamut. LMS, XYZ, Lab, DIN... these don’t clip anything at all.

I’ll try to expand the DIN99 test to a full test suite for all DIN-related tests.

The implicit assumption that RGB colors are limited to sRGB is a problem. We already saw how this caused problems this in the context of ```distinguishable_colors```, issue #241 .

I’m already thinking about ways to deal with some assumptions (whitepoints and sRGB) and how they are implemented, but I don’t have a good idea yet.

timholy: We can definitely tackle the problem in pieces. The only thing this PR needs is DIN tests.

m-lohmann:
@timholy: I added DIN tests, but during that I noticed another problem with DIN99d which seems to have to do with an error in the white white point adaptation flow. Lab(100,0,0) does not produce DIN99(100,0,0), which it should. The problem is not in the DIN99d conversion itself, but must lie in the whitepoint adaptation handling (or the lack thereof). I’m looking into that. I’ll let you know once I know more.

timholy:
Bump.

timholy:
Just as a note, I'm considering removing the DIN types from ColorTypes unless we get a fix for this issue.

PR #268, opened 2017-01-22
add example of creating colorant using hex codes

### LINKED ISSUE #267: Function or constructor to convert hex to Color object?
Is there any reason* that there isn't an inverse function for `hex`? That is, a way to convert hex triplets to a Color object. Something like:

```
Color("#7aa457")
```

or 

```
Color(ihex("#7aa457"))
```

*A reason other than no one has implemented it. I ask because there might be some problems associated with the hex codes that I'm unaware of.

tlnagy:
fixes #267


timholy:
Great, thanks!

PR #269, opened 2017-01-27
Fixed missing "=" in xml tag in Base.show

skleinbo:
Dear all,
the malformed XML tag may trigger a warning when converting the generated SVG with e.g inkscape. Conversion is successful though. Very minor issue in a rather obscure use case :)

Best,
Stephan

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -4,7 +4,7 @@
 @compat function Base.show(io::IO, ::MIME"image/svg+xml", c::Color)
     write(io,
         """
-        <?xml version"1.0" encoding="UTF-8"?>
+        <?xml version="1.0" encoding="UTF-8"?>
         <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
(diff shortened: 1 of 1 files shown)


timholy:
There's no way this can hurt anything, right? This is how it's supposed to be done? (Not my area of expertise.)

skleinbo:
No, I see no way it will break anything. Keys and values ought to be separated by an equal sign. 
For example 
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
 "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" version="1.1"
     width="25mm" height="25mm" viewBox="0 0 1 1">
     <rect width="1" height="1"
           fill="#000000" stroke="none"/>
</svg>
```
validates as SVG 1.1 markup with [https://validator.w3.org/check](https://validator.w3.org/check)

timholy:
Thanks very much, @skleinbo.

PR #272, opened 2017-02-04
Add promotions between Gray and multi-component Colorants

timholy:
Amazing that we haven't noticed the lack of these previously. These rules aren't perfect, but I'll file an issue detailing what needs to improve.

@tkelman, I also noticed some stale (>1 year old) `@deprecate` statements, which I deleted. Does deleting those mean we need to do a minor version bump?

### DIFF
--- a/src/Colors.jl
+++ b/src/Colors.jl
@@ -34,19 +34,4 @@ include("colormaps.jl")
 include("display.jl")
 include("colormatch.jl")
 
-@deprecate rgba RGBA
-@deprecate hsva HSVA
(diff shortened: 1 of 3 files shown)


tkelman:
That would be the more cautious option, yeah. Any idea if people had been using the old names in packages that weren't updated?

timholy:
I have 26 of the 52 Colors dependents installed. grepping all the `*.jl` files for `rgb(` did not reveal anything besides a lot of false positives. Good enough for me.

Issue #277, opened 2017-03-04
Some color names are missing


kimikage:
The following parsing of colors fails because `color_names` is missing some color names in 
[SVG color keywords][SVG].  

```
colorant"aqua"   # "aqua"   => (  0, 255, 255),
colorant"indigo" # "indigo" => ( 75,   0, 130),
colorant"lime"   # "lime"   => (  0, 255,   0),
colorant"teal"   # "teal"   => (  0, 128, 128),
```

[SVG]:https://www.w3.org/TR/SVG/types.html#ColorKeywords

oxinabox:
Confirmed (on 0.7.3, and on master)

As a work around:
if you use [NamedColors.jl](https://github.com/JuliaGraphics/NamedColors.jl)
They will parse.
NamedColors adds a fallback on `colorant` to its own much more extensive colors lists

Though I think the definition of Indigo disagrees.
This maybe because the there are multiple conflicting definitions for indigo.



```julia
julia> using NamedColors
julia> colorant"aqua"
RGB24{N0f8}(0.0,1.0,1.0)

julia> colorant"indigo"
RGB24{N0f8}(0.22,0.008,0.51)

julia> colorant"lime"
RGB24{N0f8}(0.0,1.0,0.0)

julia> colorant"teal"
RGB24{N0f8}(0.0,0.502,0.502)
```

kimikage:
Thank you for your confirmation.

The result of indigo in NamedColors.jl is seemed to be from the XKCD color survey.

As `colorant"red"` is parsed as `RGB{N0f8}(1.0,0.0,0.0)`, which disagrees with the XKCD's one, there is, I think, little reason to prefer the XKCD's definition.

oxinabox:
`NamedColors.jl` always prefers `Color.jl`'s definitions in preference to all others, no mater what you do.
After that point there is a configurable list of precedence for the various color dictionaries.
On which XKCD is supposed to be by default near the bottom.
I'll have to check that is working right: https://github.com/JuliaGraphics/NamedColors.jl/issues/8
thanks

timholy:
Fixed by #279

PR #279, opened 2017-03-09
Add missing color names (#277)

kimikage:
fix error on parsing "aqua", "indigo", "lime" and "teal", which are defined in SVG


timholy:
Thanks!

PR #284, opened 2017-04-06
more faithful preservation of aspect ratio when displaying as SVG

### LINKED ISSUE #283: Better aspect ratio preservation when displaying as SVG
I've noticed some minor weirdness with the aspect ratio of arrays of Colors when displayed as SVG, and I think that some improvement might be possible. The relevant code is here: https://github.com/JuliaGraphics/Colors.jl/blob/master/src/display.jl#L22

The specific set of checks in that code can lead to some very strange edge cases. For example, try creating the following arrays:

```julia
A = zeros(Gray, 9, 10)
B = zeros(Gray, 9, 18)
C = zeros(Gray, 9, 19)
```

Matrix A renders each color as a perfect square, which is great. Matrix B squashes the aspect ratio slightly, so each rectangle is a bit taller than it is wide. Matrix C, however, results in this:

screen shot 2017-04-06 at 1 40 26 pm

That is, the resulting array is taller than it is wide, and each rectangle has been dramatically stretched. That seems like the wrong behavior. 

I realize that this is a weird edge case, but it happened to come up in my work, so I figured I'd report it. 

If I may, I'd propose something like the following:

1. compute the natural aspect ratio of the array (`m` / `n`)
1. clamp that aspect ratio to some pre-determined range of values
1. size the rectangles based on that clamped aspect ratio
1. scale the rectangles to obey the min. and max. limits on the output size

If that seems reasonable, I'm happy to work on this myself and submit a PR. 

BTW @Evizero you mentioned something over in https://github.com/JuliaImages/Images.jl/issues/619 about new code to render an image in a table. Is that a proposed replacement for the SVG code here?

rdeits:
Fixes #283 

This rewrites the logic to choose the rectangle size for each color when rendering to SVG. The result should be that: 

* a much wider range of aspect ratios are faithfully preserved (the specific range is 0.16 to 7.2, which is a consequence of the particular max width, max height, and max pixel size chosen)
* aspect ratios outside that range are clamped to the nearest aspect ratio in the range
* color matrices with `m` > `n` are always drawn taller than they are wide, and vice versa
* a color matrix with `m + 1` rows is never shorter than a color matrix with `m` rows, and a color matrix with `n + 1` columns is never thinner than a color matrix with `n` columns

Here's the particular motivating example from #283 after this change:

screen shot 2017-04-06 at 3 11 07 pm

I also made another minor change: previously, there was one `show()` method for vectors and matrices, which required checking ndim() in a couple of places. I replaced that with `show(v::Vector) = show(reshape(v, (1, length(v))))` since `reshape()` doesn't copy in Julia >= v0.5. 

I would suggest playing around with this for a variety of sizes to see how it behaves. I found the following snippet useful in my testing (inside IJulia):

```julia
using Colors, Interact
c = distinguishable_colors(40 * 60)

@manipulate for m in 1:40, n in 1:60
    reshape(@view(c[1:(m * n)]), m, n)
end
```

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -1,6 +1,10 @@
 # Displaying color swatches
 # -------------------------
 
+const max_width = 180
+const max_height = 150
+const max_pixel_size = 25
+
 function Base.show(io::IO, ::MIME"image/svg+xml", c::Color)
     write(io,
         """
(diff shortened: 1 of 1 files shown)


timholy:
This is great, thanks!

PR #288, opened 2017-07-16
Fix vector construction

YingboMa:
This will fix the bug

```
julia> Colors.colormatch(359.1)
ERROR: DimensionMismatch("dimensions must match")
...
```


ararslan:
Nice, thanks! Could you add a test for this?

PR #289, opened 2017-07-26
fix depwarns under 0.7 and bump REQUIRE and CI to 0.6

Sacha0:
Best! :)


Sacha0:
Thanks Tim! :)

timholy:
Thank you! This is very helpful indeed.

PR #295, opened 2017-09-23
Use isapprox in tests

timholy:
This fixes some deprecation warnings


timholy:
Seems like an HDF5 problem (Linux passes locally on both 0.6 and 0.7). Hilarious that it's now the OSX build that is behaving nicely.

PR #304, opened 2018-06-20
fixes for julia 0.7

fredrikekre:


fredrikekre:
should wait for https://github.com/JuliaGraphics/ColorTypes.jl/pull/105

PR #312, opened 2018-08-10
Better v1.0 updates

staticfloat:
This improves on Jeff's PR, and passes tests locally.  Let's see what CI says.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
     - linux
 julia:
     - 0.7
+    - 1.0
     - nightly
 notifications:
     email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,6 @@ jobs:
       script:
         - julia -e 'using Pkg; Pkg.clone(pwd()); Pkg.build("Colors")'
         - julia -e 'using Pkg; Pkg.add("Documenter")'
-        - if [ $TRAVIS_JULIA_VERSION = "nightly" ]; then julia -e 'import Colors; cd(joinpath(pathof(Colors),"..","..")); include("docs/make.jl")'; else julia -e 'cd(Pkg.dir("Colors")); include("docs/make.jl")'; fi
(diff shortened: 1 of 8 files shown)


staticfloat:
We'll re-run tests after `JLD2` tag is available

timholy:
Anyone know who has rights on the Appveyor tests?

timholy:
Hmm, Travis failures seem related to the changes here?

.travis.yml
```diff
@@ -9,15 +9,15 @@ notifications:
     email: false
 script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-    - julia -e 'Pkg.clone(pwd()); Pkg.build("Colors"); Pkg.test("Colors"; coverage=true)';
+    - julia -e 'using Pkg; Pkg.clone(pwd()); Pkg.build("Colors"); Pkg.test("Colors"; coverage=true)';
```
fredrikekre: Is there a reason not to use the default script?

staticfloat: > Is there a reason not to use the default script?

Not as far as I can tell, I was just keeping things as similar to the previous state as possible.

Project.toml
```diff
@@ -0,0 +1,13 @@
+name = "Colors"
+uuid = "5ae59095-9a9b-59fe-a467-6f913c188581"
+authors = ["staticfloat@gmail.com <staticfloat@gmail.com>"]
```
fredrikekre: Do you want/should you have your email here?

fredrikekre:
`AssertionError: abs(dexpect - colordiff(Lab(a...), Lab(b...), metric)) < eps_cdiff` on AV. Does `eps_cdiff` need to be bigger here?

staticfloat:
I don't know this package well enough to answer.  Someone else who does (@timholy?) would have to answer.

timholy:
It's a 32-bit problem, probably getting overflow. On it.

PR #321, opened 2018-08-20
Fix version dependency on ColorTypes

timholy:
CC @juliohm. This one will be more painful because we also have to correct METADATA for the 0.9.3 release.


timholy:
https://github.com/JuliaLang/METADATA.jl/pull/17099

PR #323, opened 2018-08-23
Temporarily disable Project.toml

### LINKED ISSUE #322: Please consider deleting Project.toml
It is not yet needed as `REQUIRE` is the basis for the registry and, for the time being, the Project.toml file breaks CIBot for all packages that depend on `Colors`. That will be fixed eventually but meanwhile, it would be great if you could remove the project file.

fredrikekre:
Temporarily disable Project.toml to simplify life for JuliaCIBot when running METADATA tests, fix #322.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,17 +8,15 @@ julia:
     - nightly
 notifications:
     email: false
-script:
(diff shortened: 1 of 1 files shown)


fredrikekre:
I assume it has to be tagged if it is going to be useful for CIBot since that checks releases, @andreasnoack?

PR #325, opened 2018-09-01
Update namedcolors.md

### LINKED ISSUE #324: Named colors shouldn't be Tuple{Int64,Int64,Int64}
Hello,

I noticed

```julia
julia> Colors.color_names["indianred"]
(205, 92, 92)

julia> typeof(Colors.color_names["indianred"])
Tuple{Int64,Int64,Int64}
```

that's odd... I wonder why it is not `Tuple{UInt8,UInt8,UInt8}`

maybe a better approach could be to have `RGB{N0f8}` as value in `color_names` dictionnary.

Or provide function to get color as `RGB{N0f8}`

```julia
using Colors
using ColorTypes
using FixedPointNumbers


struct NamedColor
    s::String
end

function RGB{T}(namedcolor::NamedColor) where {T}
    RGB(map(c->reinterpret(T, c), UInt8.(Colors.color_names[namedcolor.s]))...)
end

RGB(namedcolor::NamedColor) = RGB{N0f8}(namedcolor)

namedcolor = NamedColor("indianred")

println(RGB{N0f8}(namedcolor))
println(RGB(namedcolor))
```

Maybe having some other function to return any `RGB{T}` from `NamedColor` should be considered.

Kind regards

scls19fr:
Closes #324

### DIFF
--- a/docs/src/namedcolors.md
+++ b/docs/src/namedcolors.md
@@ -10,12 +10,12 @@ color_names = Dict(
     ...
 ```
 
-Named colors are available using:
+Named colors are available as `RGB{N0f8}` using:
 
 ```julia
 julia> using Colors
 
-julia> colorant"indianred"
+julia> color = colorant"indianred"
 RGB{N0f8}(0.804,0.361,0.361)
 ```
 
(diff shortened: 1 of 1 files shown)


timholy:
Thanks!

PR #327, opened 2018-09-13
Misc fixes

fredrikekre:
Misc fixes
 - Enable `Project.toml` again
 - Use `docs/Project.toml` to build docs
 - Build docs from Julia v1.0
 - Move `Test` and `JLD2` to test-only dependencies
 - Remove unused `test/REQUIRE`


docs/Project.toml
```diff
@@ -0,0 +1,5 @@
+[deps]
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+
+[compat]
+Documenter = "~0.19"
```
fredrikekre: Adding this too, since I happen to know that Documenter v0.20 will be breaking. When v0.20 is released and Colors uses the new stuff in Documenter this `[compat]` can be removed or, perhaps better, updated to
```
Documenter "^0.20"
```

fredrikekre:
If this gets merged soonish I can point to Colors as a nice example on how to build docs in Julia v1.0 :smile:

PR #329, opened 2018-10-23
Use ReshapedArray for show()

### LINKED ISSUE #328: Fix show() method to reshape a copy
Per discussion in Slack with @pfitzseb in #juno-bridged on slack, there's an issue with the `show()` method that causes weird errors in Juno. He suggested that [this line](https://github.com/JuliaGraphics/Colors.jl/blob/bc7221505b0f70377c8cb1c9cd6ac522ae0c240b/src/display.jl#L24) should instead be:

```
show(io, MIME"image/svg+xml"(), reshape(copy(cs), (1, length(cs))))
```

kescobo:
Fixes #328. Two questions:

1) I couldn't find the show method in tests - should I add one? If so, where makes the most sense?
2) Were you suggesting that I *add* the `transpose()` method for colors, or is this sufficient?

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -19,9 +19,9 @@ function Base.show(io::IO, ::MIME"image/svg+xml", c::Color)
         """)
 end
 
-function Base.show(io::IO, ::MIME"image/svg+xml", 
+function Base.show(io::IO, ::MIME"image/svg+xml",
                    cs::AbstractVector{T}) where T <: Color
-    show(io, MIME"image/svg+xml"(), reshape(cs, (1, length(cs))))
+    show(io, MIME"image/svg+xml"(), Base.ReshapedArray(cs, (1, length(cs)), ()))
 end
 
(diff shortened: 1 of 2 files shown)


timholy:
> I couldn't find the show method in tests - should I add one

This package predated Julia's testing infrastructure :smile:. Yeah, would be good to add it to tests, including the grow-the-array test that triggered this change. Perhaps to `colormaps.jl`.

> I add the transpose() method

Nope, this is right. In general we don't want to support much math on colors here, and transpose probably shouldn't be defined.

kescobo:
Alright - I added a simple test. I'm not sure if there's a way to test it more directly - if it doesn't work the error happens outside a test, but it at least checks that the result of the `push!()` succeeds. 

I verified that the code in the test *doesn't* work on the current master, but does work on this branch.

timholy:
Awesome, thanks @kescobo!

kescobo:
No sweat!

Issue #334, opened 2019-06-27
Approx wavelength to RGB converter [Suggested addition]


ianshmean:
The conversion of wavelength to color on a RGB plane is a complicated can of worms.. but I needed something crude, so converted the code below. Perhaps it's useful to add to this package?

```julia
using ColorTypes
function wavelength_to_rgb(wavelength;gamma=0.8)
    #=
    Based on code by Dan Bruton http://www.physics.sfasu.edu/astro/color/spectra.html
    discussed: https://stackoverflow.com/questions/1472514/convert-light-frequency-to-rgb
    python version: https://gist.github.com/friendly/67a7df339aa999e2bcfcfec88311abfc
    
    =#
    if (wavelength >= 380) && (wavelength <= 440)
        attenuation = 0.3 + 0.7 * (wavelength - 380) / (440 - 380)
        R = ((-(wavelength - 440) / (440 - 380)) * attenuation) ^ gamma
        G = 0.0
        B = (1.0 * attenuation) ^ gamma
    elseif (wavelength >= 440) && (wavelength <= 490)
        R = 0.0
        G = ((wavelength - 440) / (490 - 440)) ^ gamma
        B = 1.0
    elseif (wavelength >= 490) && (wavelength <= 510)
        R = 0.0
        G = 1.0
        B = (-(wavelength - 510) / (510 - 490)) ^ gamma
    elseif (wavelength >= 510) && (wavelength <= 580)
        R = ((wavelength - 510) / (580 - 510)) ^ gamma
        G = 1.0
        B = 0.0
    elseif (wavelength >= 580) && (wavelength <= 645)
        R = 1.0
        G = (-(wavelength - 645) / (645 - 580)) ^ gamma
        B = 0.0
    elseif (wavelength >= 645) && (wavelength <= 750)
        attenuation = 0.3 + 0.7 * (750 - wavelength) / (750 - 645)
        R = (1.0 * attenuation) ^ gamma
        G = 0.0
        B = 0.0
    else
        R = 0.0
        G = 0.0
        B = 0.0
    end
    return RGB(R,G,B)
end
```
```julia
using Colors
wavelength_to_rgb.(320:1:800)
```
<img width="691" alt="Screen Shot 2019-06-27 at 2 11 15 PM" src="https://user-images.githubusercontent.com/1694067/60289877-776bd880-98e5-11e9-99ed-4d85a2aade32.png">

cormullion:
[...]

PR #335, opened 2019-06-28
Update docs to highlight wavelength to RGB conversion

ianshmean:
As discussed in #334 

Based on https://gist.github.com/friendly/67a7df339aa999e2bcfcfec88311abfc
Which is a conversion of code by Dan Bruton http://www.physics.sfasu.edu/astro/color/spectra.html
And discussed: https://stackoverflow.com/questions/1472514/convert-light-frequency-to-rgb


ianshmean:
I also corrected the version in `Project.toml` and deleted `REQUIRE`

ianshmean:
Closing as it turns out a more accurate method is already provided via
```julia
using Colors
RGB.(colormatch.(320:800))
```

cormullion:
How about turning this into a documentation PR? It’s clearly needed. 😀

ianshmean:
@cormullion Good idea. I've added a short example and figure

PR #337, opened 2019-07-17
Support range(start, stop; kwargs...)

crstnbr:
From Julia 1.1 on, `range(start, stop; kwargs...)` is supported complementary to `range(start; stop, kwargs...)`. This PR adds the former method for `Colorant`s.

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -58,7 +58,9 @@ function range(start::T; stop::T, length::Integer=100) where T<:Colorant
     return T[weighted_color_mean(w1, start, stop) for w1 in range(1.0,stop=0.0,length=length)]
 end
 
-range(start::T, stop::T; kwargs...) where T<:Colorant = range(start; stop=stop, kwargs...)
+if VERSION >= v"1.1"
+    range(start::T, stop::T; kwargs...) where T<:Colorant = range(start; stop=stop, kwargs...)
+end
 
 if VERSION < v"1.0.0-"
 import Base: linspace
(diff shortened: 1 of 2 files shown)


crstnbr:
bump

src/utilities.jl
```diff
@@ -58,6 +58,8 @@ function range(start::T; stop::T, length::Integer=100) where T<:Colorant
     return T[weighted_color_mean(w1, start, stop) for w1 in range(1.0,stop=0.0,length=length)]
 end
 
+range(start::T, stop::T; kwargs...) where T<:Colorant = range(start; stop=stop, kwargs...)
```
johnnychen94: Perhaps you need to wrap it inside an if block

```julia
if VERSION >= v"1.1"
	...
end
```

crstnbr:
Good to go?

johnnychen94:
It looks good to me. Without this patch, it would raise some strange error message.

```julia
julia> range(Gray(0), Gray(1), length=11)
11-element LinRange{Gray{Normed{UInt8,8}}}:
Error showing value of type LinRange{Gray{Normed{UInt8,8}}}:
ERROR: MethodError: no method matching *(::Float64, ::Gray{Normed{UInt8,8}})
Closest candidates are:
  *(::Any, ::Any, ::Any, ::Any...) at operators.jl:529
  *(::Float64, ::Float64) at float.jl:399
  *(::AbstractFloat, ::Bool) at bool.jl:112
  ...

```

Unfortunately, @timholy is quite busy on his daily work recently; you might need to wait for a while until he finds some time.

crstnbr:
Alright, no worries. Thanks for the info!

Issue #278, opened 2017-03-08
Bradford transformation support


kimikage:
As `README.md` says, Colors.jl already has a useful function for chromatic adaptation named `whitebalance`.  
`whitebalance` uses the CAT02 transformation matrix to convert colors into LMS.  
However, not only the CAT02 but also the linear Bradford transformation is commonly used in the field of color management because of the interoperability.  
In fact, to minimize the risk of conflict, the Bradford model is recommended in the [ICC specification](http://www.color.org/specification/ICC1v43_2010-12.pdf).

We can obtain the adaptation result in the Bradford model by converting the colors into LMS with the Bradford transformation in advance of calling `whitebalance`, and converting the result back into the original color space.
I think it's too much bother to do that every time, though.

A simple solution is to add a new function such as the following:

```julia
# the Bradford transformation matrix from v4 ICC Specification
const BFD = [ 0.8951  0.2664 -0.1614
             -0.7502  1.7135  0.0367
              0.0389 -0.0685  1.0296 ]
const BFD_INV = inv(BFD)

function bradford{T <: Color}(c::T, src_white::Color, ref_white::Color)
    c_xyz = convert(XYZ, c)
    src_xyz = convert(XYZ, src_white)
    dest_xyz = convert(XYZ, ref_white)
    
    c_lms = BFD * [c_xyz.x, c_xyz.y, c_xyz.z]
    src_wp = BFD * [src_xyz.x, src_xyz.y, src_xyz.z]
    dest_wp = BFD * [dest_xyz.x, dest_xyz.y, dest_xyz.z]
    ans = BFD_INV * (c_lms .* dest_wp ./ src_wp)

    convert(T, XYZ(ans[1], ans[2], ans[3]))
end
```
But I don't know whether it makes sense to propose such an ad-hoc function.  

Another solution is to make changes to `cnvt{T}(::Type{LMS{T}}, c::XYZ)` and `whitebalance` to take the CAT matrix as an optional argument.  

Do you have any ideas?

m-lohmann:
I have been thinking about implementing the handful of most important CATs a while ago, including the BFD (linear or non-linear), simple vonKries etc.

As far as I remember CAT02 is basically the BFD, but I might be wrong.

The problem is that there is no strict color conversion pipeline enforced in the whole package.
Another problem is that RGB is not strictly enforced to be a device color space with potentially negatve values for later processing etc.
[...]

PR #340, opened 2019-08-26
[WIP]  Add new chromatic adaptation function `adaptation` (#278)

kimikage:
This is a solution for #278.

TODO: 
- Discuss the name and interface
- Add tests and documents
- (Add some CAT new?)

### DIFF
--- a/src/adaptation.jl
+++ b/src/adaptation.jl
@@ -188,7 +188,7 @@ Returns:
 function adaptation(c::T;
                     src_white::Color=WP_DEFAULT,
                     ref_white::Color=WP_DEFAULT,
-                    cat::AbstractCAT=CAT02()) where T <: Color
+                    cat::AbstractCAT=CAT_02()) where T <: Color
     convert(T, adaptation_xyz(c, src_white, ref_white, cat), ref_white)
 end
 
--- a/src/adaptation.jl
+++ b/src/adaptation.jl
@@ -1,0 +1,128 @@
+# Chromatic Adaptation / Whitebalancing
+# -------------------------------------
+
+# Define an abstract type to represent chromatic adaptation transform
+abstract type CAT end
+
+"""
+There is incompatibility between `MatrixCAT` and `NonLinearCAT`.
+In other words, `MatrixCAT` is never a special case or subtype of `NonLinearCAT`.
+
+In the `MatrixCAT` adaptation, the flow is as follows:
+    `XYZ` -(Matrix CAT)-> `LMS` -(white point conversion)-> `LMS` -(inv. Matrix CAT)-> `XYZ`
+In the white point conversion, the von Kries model, i.e. the diagonal matrix, is used.
+
+
+In the `NonLinearCAT` adaptation, the flow is as follows:
+    `XYZ` -(CAT w/ white balancing)-> `LMS` -(inv. CAT w/ white balancing)-> `XYZ`
+
+"""
+abstract type MatrixCAT <: CAT end
+
+abstract type NonLinearCAT<: CAT end
+
+const EYE3x3 = [1.0 0.0 0.0; 0.0 1.0 0.0; 0.0 0.0 1.0]
+
+# Hunt-Pointer-Estevez (also called von Kries) transformation matrix
+const HPE = [ 0.38971  0.68898 -0.07868
+             -0.22981  1.18340  0.04641
+              0.00000  0.00000  1.00000 ]
+
+const HPE_INV = inv(HPE)
+
+# Bradford transformation matrix from v4 ICC Specification
+const BFD = [ 0.8951  0.2664 -0.1614
+             -0.7502  1.7135  0.0367
+              0.0389 -0.0685  1.0296 ]
+
+const BFD_INV = inv(BFD)
+
+# Chromatic adaptation from CIECAM97s
+const CAT97s = [ 0.8562  0.3372 -0.1934
+                -0.8360  1.8327  0.0033
+                 0.0357 -0.0469  1.0112 ]
+
+const CAT97s_INV = inv(CAT97s)
+
+# Chromatic adaptation from CIECAM02
+const CAT02 = [ 0.7328 0.4296 -0.1624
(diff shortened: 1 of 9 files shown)


src/adaptation.jl
```diff
@@ -0,0 +1,231 @@
...
+                        cat::AbstractCAT=CAT_02())
+    # this is the wrapper method
+
+    adaptation_xyz(c, src_white, ref_white, cat)
+end
+
+function adaptation_xyz(c::Color, src_white::Color, ref_white::Color, cat::MatrixCAT)
+    c_lms = convert(LMS, c, src_white, cat=cat)
+    src_wp = convert(LMS, src_white, src_white, cat=cat)
+    ref_wp = convert(LMS, ref_white, src_white, cat=cat)
```
kimikage: `ref_white` should be converted to LMS with **`src_white`** here, right?

However, in NonLinearCATs' process, it seems natural that `ref_white` is converted to LMS with **`ref_white`**.
Of course, this is a trivial matter because `ref_white` and `src_white` are usually in XYZ.

src/adaptation.jl
```diff
@@ -0,0 +1,231 @@
...
+end
+
+
+
+# Define "domain-specific" conversion methods to simplify the conversion chains
+
+# xyz_to_lms()
+# ------------
+function xyz_to_lms(c::XYZ{T}, ::XYZ, cat::MatrixCAT) where T
+    @mul3x3 LMS{T} cat_cnvt(cat) c.x c.y c.z
```
kimikage: Oops
```suggestion
    @mul3x3xyz LMS{T} cat_cnvt(cat) c
```

src/adaptation.jl
```diff
@@ -0,0 +1,231 @@
...
+Returns:
+  A whitebalanced color.
+"""
+function whitebalance(c::T, src_white::Color, ref_white::Color) where T <: Color
+    c_lms = convert(LMS, c) # this does not take account of `src_white`
+    src_wp = convert(LMS, src_white)
+    ref_wp = convert(LMS, ref_white)
+
+    # This is sort of simplistic, it sets the degree of adaptation term in
+    # CAT02 to 1.
```
kimikage: The original comment was **"to 0"**(cf. `src/algorithms.jl`), but I think this simplification assumes D=1.

src/conversions.jl
```diff
@@ -271,6 +267,12 @@ cnvt(::Type{HSI{T}}, c::Color3) where {T} = cnvt(HSI{T}, convert(RGB{T}, c))
 # Everything to XYZ
 # -----------------
 
+convert(::Type{XYZ}, c, wp::XYZ) = convert(XYZ{eltype(wp)}, c, wp)
```
kimikage: Although `XYZ{eltype(wp)}` is not so natural and incompatible with the variant method without `wp` argument, this simple rule is practical.

src/conversions.jl
```diff
@@ -378,9 +383,16 @@ function cnvt(::Type{XYZ{T}}, c::DIN99d) where T
 
 end
 
+function convert(::Type{XYZ}, c::LMS, wp::XYZ; cat::AbstractCAT=CAT_02())
```
kimikage: The third argument `wp` is not a keyword argument because of the compatibility with `convert()`s of Lab-like types.

It is probably a good idea that all optional arguments in keyword style in the future. (We need to have deeper discussions about this. Not now.)

src/adaptation.jl
```diff
@@ -0,0 +1,231 @@
...
+- `ref_white`: Reference or destination white.
+- `cat`: Chromatic adaptation transform model.
+
+Returns:
+  A adapted color.
+"""
+function adaptation(c::T;
+                    src_white::Color=WP_DEFAULT,
+                    ref_white::Color=WP_DEFAULT,
+                    cat::AbstractCAT=CAT02()) where T <: Color
```
kimikage: Terrible typo!
Since the const `CAT02` is not exported, the same mistake will not be repeated in user codes.
```suggestion
                    cat::AbstractCAT=CAT_02()) where T <: Color
```

src/adaptation.jl
```diff
@@ -0,0 +1,231 @@
...
+    lms_to_xyz(ans, ref_white, cat)
+end
+
+function adaptation_xyz(c::Color, src_white::Color, ref_white::Color, cat::NonLinearCAT)
+    ans = convert(LMS, c, src_white, cat=cat)
+    lms_to_xyz(ans, ref_white, cat)
+end
+
+"""
+    adaptation(c; src_white, ref_white, cat::AbstractCAT=CAT02())
```
kimikage:
```suggestion
    adaptation(c; src_white, ref_white, cat::AbstractCAT=CAT_02())
```

src/conversions.jl
```diff
@@ -53,20 +53,14 @@ ColorTypes._convert(::Type{Cdest}, ::Type{Odest}, ::Type{Osrc}, c) where {Cdest<
 cnvt(::Type{C}, c::TransparentColor) where {C<:Color} = cnvt(C, color(c))
 cnvt(::Type{C}, c) where {C} = convert(C, c)
 
+# Fallback to catch operations which do not depend on the white point
+convert(::Type{C}, c, wp::XYZ) where {C<:Color} = cnvt(C, c)
```
kimikage: This may stir up a hornets nest.
The current `RGB` means "sRGB" (D65) regardless of the white point.
Because there is a lot of discussion, I'm going to pick out the enhanced support for white points as an independent PR.

timholy:
I'd love it if someone who knows more about color spaces than me reviewed this.

kimikage:
No problem. I think there is no need to rush this. 
However, the merging becomes difficult over time, so I'd like to subdivide this into several PRs.

timholy:
That's fine. You should also feel free to merge this work when you wish. I don't see anything that bothers me, and I'm especially grateful for `@mul3x3` and friends. The tricky parts go well beyond my expertise to evaluate properly.

Issue #341, opened 2019-08-31
`show()` may output negative size swatches


kimikage:
There are cases where `show()` outputs the SVG swatches with negative sizes as follows:
```julia
julia> buf = IOBuffer();

julia> show(buf, "image/svg+xml", rand(RGB, 28, 181));

julia> seekstart(buf); for i=1:9; println(readline(buf)) end
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
 "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" version="1.1"
     width="179.99999999999997mm" height="27.845303867403313mm"
     shape-rendering="crispEdges">
<rect x="0.0mm" y="0.0mm"
      width="0.9944751381215469mm" height="-0.005524861878453136mm"
      fill="#B06692" stroke="none" />

julia> buf = IOBuffer();

julia> show(buf, "image/svg+xml", rand(RGB, 151, 28));

julia> seekstart(buf); for i=1:9; println(readline(buf)) end
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
 "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" version="1.1"
     width="27.814569536423843mm" height="150.0mm"
     shape-rendering="crispEdges">
<rect x="0.0mm" y="0.0mm"
      width="-0.0066225165562913135mm" height="0.9933774834437087mm"
      fill="#2973DF" stroke="none" />
```

Since the negative width and height are not allowed in SVG, the renderer may report errors and no swatch is rendered.

The cause is insufficient consideration for the scaling (i.e. actual `xsize`/`ysize`) in the calculation of the padding sizes.
https://github.com/JuliaGraphics/Colors.jl/blob/657e265a4cccc9fc32584f7f3505611537bfc79d/src/display.jl#L37-L41

Although this is a trivial bug, we should fix it before tackling the issue #317.

timholy:
Fixed by #342

Issue #317, opened 2018-08-14
Limit for displaying vector of colors in IJulia


mprat:
When trying to display large vectors of Colors (for example when you have a long row of pixels from Images.jl) in IJulia, there should be a default limit for when the display gets rendered as a PNG rather than an SVG. For large vectors and slow computers (or computers with not that much RAM), displaying a 1000-point vector in an SVG in IJulia bricks the computer.

kimikage:
The following is not a drastic solution but one of the mitigation measures:
https://github.com/JuliaGraphics/Colors.jl/compare/master...kimikage:display

It is easy to limit the matrix size.
However, I don't think it is a simple matter of what should be displayed.
Top left corner? Four corners?
Decimation?

By the way, I found a little [bug](https://github.com/JuliaGraphics/Colors.jl/commit/dacf8578083bf90d3653cde661106526e0a616b4).

kimikage:
Related issues:
https://github.com/JuliaLang/IJulia.jl/issues/229
https://github.com/JuliaImages/ImageShow.jl/issues/11

kimikage:
Another reduction method was implemented: https://github.com/JuliaGraphics/Colors.jl/pull/346#issuecomment-529076468

PR #342, opened 2019-08-31
Fix negative size swatches in output of show() (#341)

kimikage:
Fix negative size error(#341), Simplify SVG output and Add tests


timholy:
Very nice @kimikage! Let me know if you want me to tag this with a new release, or if you'd prefer to incorporate more changes first.

kimikage:
Well, I should probably send the PRs together for the `show()`-related issues #317 and #343.
Then, I respect your decision.

PR #344, opened 2019-09-04
Several fixes

timholy:
This does some long needed-maintenance. Most importantly, I finally just took the DIN99 changes despite the lack of good tests. (Would be really good to have those...)

### DIFF
--- a/src/colormaps.jl
+++ b/src/colormaps.jl
@@ -242,7 +242,7 @@ end
 # Main function to handle different predefined colormaps
 #
 """
-    colormap(cname, [N; mid, logscale, kvs...])
+    colormap(cname, N=100; mid=0.5, logscale=false, kvs...])
 
 Returns a predefined sequential or diverging colormap computed using
 the algorithm by Wijffelaars, M., et al. (2008).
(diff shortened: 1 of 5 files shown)


src/colormaps.jl
```diff
@@ -5,7 +5,12 @@ include("maps_data.jl")
 # ----------------------
 
 """
-    distinguishable_colors(n, [seed]; [transform, lchoices, cchoices, hchoices])
+    colors = distinguishable_colors(n, seed=RGB{N0f8}[];
+                                    dropseed=false,
+                                    transform=identity,
+                                    lchoices=range(0, stop=100, length=15),
+                                    cchoices=range(0, stop=100, length=15),
+                                    hchoices=range(0, stop=340, length=20))
```
kimikage: This is off the topic, but `hchoices=range(0, stop=340, length=20)` is a little strange.
`hchoices=range(0, stop=342, length=20)` (or `hchoices=range(0, stop=340, length=18)` ) ?

src/conversions.jl
```diff
@@ -354,23 +354,22 @@ cnvt(::Type{XYZ{T}}, c::LCHuv) where {T} = cnvt(XYZ{T}, convert(Luv{T}, c))
 function cnvt(::Type{XYZ{T}}, c::DIN99d) where T
 
     # Go back to C-h space
-    # FIXME: Clean this up (why is there no atand?)
-    h = rad2deg(atan(c.b,c.a)) + 50
+    h = atand(c.b, c.a) - 50
     while h > 360; h -= 360; end
     while h < 0;   h += 360; end
```
kimikage: `h` is used later only for `sind(h)` and  `cosd(h)`.
This normalization seems unnecessary.

PR #346, opened 2019-09-04
Use stroke instead of fill (rect) with swatch reduction for large matrix

kimikage:
This reduces the size of SVG document and the rendering time.
So this is one way to mitigate the issue #317.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -4,7 +4,7 @@
 const max_width = 180
 const max_height = 150
 const max_swatch_size = 25
-const max_num_of_swatches = 128 * 128
+const default_max_swatches = 128 * 128
 
 function Base.show(io::IO, mime::MIME"image/svg+xml", c::Color)
     write_declaration(io, mime)
@@ -18,14 +18,19 @@ function Base.show(io::IO, mime::MIME"image/svg+xml", c::Color)
     flush(io) # return nothing
 end
 
-function Base.show(io::IO, mime::MIME"image/svg+xml",
-                   cs::AbstractVector{T}) where T <: Color
-    show(io, mime, Base.ReshapedArray(cs, (1, length(cs)), ()))
+function Base.show(io::IO, mime::MIME"image/svg+xml", cs::AbstractVector{T};
+                   max_swatches::Integer=0) where T <: Color
+    mat = Base.ReshapedArray(cs, (1, length(cs)), ())
+    show(io, mime, mat, max_swatches=max_swatches)
 end
 
-function Base.show(io::IO, mime::MIME"image/svg+xml",
-                           cs::AbstractMatrix{T}) where T <: Color
+function Base.show(io::IO, mime::MIME"image/svg+xml", cs::AbstractMatrix{T};
+                   max_swatches::Integer=0) where T <: Color
     m, n = size(cs)
+    actual_max_swatches = max_swatches > 0 ? max_swatches : default_max_swatches
+    if m * n > actual_max_swatches
+        return show_strokes(io, mime, cs, max_swatches=max_swatches)
+    end
     w = min(n * max_swatch_size, max_width)
     h = min(m * max_swatch_size, max_height)
     if max_width * m > max_height * n
@@ -38,7 +43,7 @@ function Base.show(io::IO, mime::MIME"image/svg+xml",
     sw = simplify(1 - (w / n < 3.6 ? 0 : n / w))
     sh = simplify(1 - (h / m < 3.6 ? 0 : m / h))
     if sw == "1" && sh == "1"
-        return show_strokes(io, mime, cs)
+        return show_strokes(io, mime, cs, max_swatches=max_swatches)
     end
 
     comp(x) = round(x)==round(x, digits=2) ? Int(round(x)) : round(x, digits=2)
@@ -63,9 +68,17 @@ function Base.show(io::IO, mime::MIME"image/svg+xml",
     flush(io) # return nothing
 end
 
-function show_strokes(io::IO, mime::MIME"image/svg+xml",
-                      cs::AbstractMatrix{T}) where T <: Color
+function show_strokes(io::IO, mime::MIME"image/svg+xml", cs::AbstractMatrix{T};
+                      max_swatches::Integer=0) where T <: Color
     m, n = size(cs)
+    actual_max_swatches = max_swatches > 0 ? max_swatches : default_max_swatches
+    # When `max_swatches` is specified, the following warning is suppressed.
+    if max_swatches == 0 && m * n > actual_max_swatches
+        @warn """
+            Output swatches are reduced due to the large size ($m×$n).
+            Load the ImageShow package for large images."""
+        yield()
+    end
     w = max_width
     h = max_height
     if max_width * m > max_height * n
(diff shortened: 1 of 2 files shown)


timholy:
How much of a speed improvement is this? It would also be nice to see what difference, if any, one notices in the rendered image.

kimikage:
Since IJulia (Jupyter) displays the output asynchronously, the measurement is not easy.
The followings are the outputs of Google Chrome's profiler, and just for reference purposes.

## 128x128 RGB Matrix
`[RGB(r/128,0.5,b/128) for r=1:128, b=1:128]`

|master (b39b9186672515f0c390271d57864f95b213ee25)| this (60e6d91df41cd49f09a752f06c70c2a8b8490160)  |
|:-:|:-:|
|474 ms|377 ms|
|fill_128x128|stroke_128x128|

## 512x512 Mandrill
`testimage("mandrill")`

|master (b39b9186672515f0c390271d57864f95b213ee25)| this (60e6d91df41cd49f09a752f06c70c2a8b8490160) w/ decimation  |
|:-:|:-:|
|9.57 s|0.37 s|
|fill_mandrill|stroke_mandrill_decimated|

|master (b39b9186672515f0c390271d57864f95b213ee25)| w/o decimation (reference) |
|:-:|:-:|
|9.57 s|7.78 s|
|fill_mandrill|stroke_mandrill_full|

kimikage:
## Decimation
fill_stroke
This is an example of the decimation (1/16 = 1/4 * 1/4).

## Differences in XML
| version          | one swatch element |
|-|-|
|v0.9.6             |```<rect x="1.171875mm" y="1.171875mm"```<br>```      width="0.29296875mm" height="0.29296875mm"```<br>```      fill="#4C5328" stroke="none" />```|
|master (#342 )|```<rect width="1" height="1" x="4" y="4" fill="#4C5328" />```|
|this (#346)<br>(for large matrix)      |```<path d="M4,5h1" stroke="#4C5328" />```|

src/display.jl
```diff
@@ -59,6 +63,40 @@ function Base.show(io::IO, mime::MIME"image/svg+xml",
...
+    write(io,
+        """
+        <svg xmlns="http://www.w3.org/2000/svg" version="1.1"
+             width="$(comp(w))mm" height="$(comp(h))mm" viewBox="0 0.5 $n $m"
+             stroke-width="1" stroke-linecap="butt"
+             shape-rendering="crispEdges">
+        """)
+
+    for i in 1:d:m, j in 1:d:n
+        c = hex(cs[i, j])
```
timholy: What about sending the mean rather than a "randomly-chosen" pixel? Even better would be to call [`restrict`](https://github.com/JuliaImages/ImageTransformations.jl/blob/master/src/resizing.jl), since it anti-aliases, but that uses a lot more machinery (unless we move it to its own package).

If you want to compute the mean you could do
```julia
csum, n = zero(floattype(T)), 0
for ii in i:min(m, i+d-1), jj in j:min(n, j+d-1)
    csum += cs[i, j]
    n += 1
end
c = hex(csum/n)
```

kimikage: Yes, the mean is nice for displaying *pictures*. I have no clear reason to refuse this.

However, I think it is out of the scope of `Colors.jl` to display raster images (although this words sound like a contradiction with my PR).
So, I don't want to show large images in `Colors.jl` with additional dependencies.

I also think the output have to suggest that it (i.e. large image) is not suitable for rendering as SVG. The comb-like strange outputs follow this manner.

Of course, it is another reasonable strategy to display pictures as *beautiful* as possible without such a hint.

A compromise plan is adding an overlay text such as "512×512 Array{RGB{N0f8},2}", although it mess up a beautiful picture.

A completely different strategy is to give up displaying (i.e. `throw()`). In that case, IJulia tries to show it in other formats. However, I think the incomplete images are somewhat more useful than the truncated texts.

What shall we do?

timholy: What about displaying as a slow, complete SVG but to issue a warning above a certain size? E.g.,
```julia
@warn "Slow display of color array (load the ImageShow package for large images)"
yield()  # to allow the warning to be issued while user waits
show_it_anyway(colormatrix)
```

kimikage: Thank you for your advice! `@warn` seems appropriate to this case.
However, IJulia (client) shows the warning unfortunately after displaying the matrix.
So, what about using `@warn` for notifying that the output is downsized?

timholy: That seems OK. Is there a way to get the full display if you're willing to wait?

kimikage: Hmm, is the huge SVG worth waiting for? What size is adequate for the threshold may be debatable, though.

kimikage:
I changed the reduction method, and did not squash the commits for future reference.
reduction

kimikage:
I also added the keyword argument `max_swatches`.
However, I don't know the elegant way for sending the additional argument to IJulia.
max_swatches

src/display.jl
```diff
@@ -4,6 +4,7 @@
 const max_width = 180
 const max_height = 150
 const max_swatch_size = 25
+const default_max_swatches = 128 * 128
```
timholy: One thought: should we make this
```julia
const default_max_swatches = Ref(128*128)
```

That would allow users to customize it in a session with `Colors.default_max_swatches[] = 32`.

src/display.jl
```diff
@@ -17,14 +18,19 @@ function Base.show(io::IO, mime::MIME"image/svg+xml", c::Color)
     flush(io) # return nothing
 end
 
-function Base.show(io::IO, mime::MIME"image/svg+xml",
-                   cs::AbstractVector{T}) where T <: Color
-    show(io, mime, Base.ReshapedArray(cs, (1, length(cs)), ()))
+function Base.show(io::IO, mime::MIME"image/svg+xml", cs::AbstractVector{T};
+                   max_swatches::Integer=0) where T <: Color
```
timholy: Should we consider using -1 to indicate "use the default"? I wonder if there's ever a case where someone might want to suppress output completely?

src/display.jl
```diff
@@ -59,6 +68,60 @@ function Base.show(io::IO, mime::MIME"image/svg+xml",
...
+        <svg xmlns="http://www.w3.org/2000/svg" version="1.1"
+             width="$(comp(w))mm" height="$(comp(h))mm"
+             viewBox="0 -$(comp(d/2)) $n $m" stroke-width="$d"
+             stroke-linecap="butt" shape-rendering="crispEdges">
+        """)
+
+    for i in 1:d:m, j in 1:d:n
+        # since there is no universal way,
+        # calculate the mean color in "RGB space"
+        csum = Float32[0, 0, 0]
```
timholy: Ah yes. My implementation implicitly assumed that one had loaded ColorVectorSpace, which is not an assumption we want to make. Even though this will likely be slower (each sum involves an array allocation), I doubt it matters.

src/display.jl
kimikage: This is simply because the current implementation does not accept 0 (divide by zero).
The *small* `max_swatches` already has some layout problems (e.g. the swatches are too large as shown above, are always square, never have padding,... ).
I think the problems are trivial and `max_swatches` cannot be a useful switch.
We have `;`, which is useful way to suppress output.

timholy:
Thanks very much @kimikage!

src/display.jl
kimikage: Just for future reference, a rough benchmark also said that it did not matter (after the second execution). It seems to be less critical than the allocations for output text.

timholy: Hopefully a future compiler version will also figure this out and not allocate the memory.

Issue #343, opened 2019-08-31
Displaying `TransparentColor` as SVG


kimikage:
`Color` and array of `Color` data can be displayed as SVG.
Is there a demand for displaying `TransparentColor`?

A `TransparentColor` (not arrays) can be displayed by the following function:
```julia
function Base.show(io::IO, mime::MIME"image/svg+xml", c::TransparentColor)
    id = "pat_" * string(rand(UInt32), base=62)
    hexc = hex(color(c))
    op = alpha(c)
    write_declaration(io, mime) # see  https://github.com/JuliaGraphics/Colors.jl/pull/342
    write(io,
        """
        <svg xmlns="http://www.w3.org/2000/svg" version="1.1"
             width="25mm" height="25mm" viewBox="0 0 1 1" stroke="none">
             <defs>
                <pattern id="$id" width=".2" height=".2"
                         patternUnits="userSpaceOnUse">
                    <path d="M.1,0h.1v.1h-.2v.1h.1z" fill="#999" opacity=".5" />
                </pattern>
             </defs>
             <rect width="1" height="1" fill="url(#$id)" />
             <path d="M1,0v1h-1z" fill="#$hexc" />
             <rect width="1" height="1" fill="#$hexc" fill-opacity="$op" />
        </svg>
        """)
    flush(io)
end
```
![transparentcolor](https://user-images.githubusercontent.com/12679384/64062420-0c11f000-cc21-11e9-9112-128f09caa568.png)

Displaying SVG matrices of `TransparentColor` requires attention to performance issues, since `ImageShow.jl` suppresses [only the SVG matrices of `Color`](https://github.com/JuliaImages/ImageShow.jl/blob/d9843bce9d7a7d3aa51dd02862d9dee3a2407626/src/showmime.jl#L13-L17).

kimikage:
The result of vector (1-D array) version looks like this:
![transparent_vector](https://user-images.githubusercontent.com/12679384/64063721-09b89180-cc33-11e9-89a9-fa24d901a968.jpg)

timholy:
I like it!

kimikage:
The latter design was implemented and merged.

PR #348, opened 2019-09-08
Add support for displaying `TransparentColor` as SVG (#343)

kimikage:
This adds support for displaying a single `TransparentColor` and vector of `TransparentColor` as SVG.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -18,11 +18,73 @@ function Base.show(io::IO, mime::MIME"image/svg+xml", c::Color)
     flush(io) # return nothing
 end
 
+function Base.show(io::IO, mime::MIME"image/svg+xml", c::TransparentColor)
+    show(io, mime, [c], max_swatches=default_max_swatches)
+end
+
+
 function Base.show(io::IO, mime::MIME"image/svg+xml", cs::AbstractVector{T};
                    max_swatches::Integer=0) where T <: Color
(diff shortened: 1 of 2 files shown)


kimikage:
I believe no one uses such a long vector, but implements the tail truncation.
tp_trunc

Other examples are shown in #343 (the latter design was chosen).

src/display.jl
```diff
@@ -108,7 +108,9 @@ function show_strokes(io::IO, mime::MIME"image/svg+xml", cs::AbstractMatrix{T};
         v = min(m-i+1, d) # cell height
         for y in i:i+v-1, x in j:j+u-1
             rgb = convert(RGB{Float32}, cs[y, x])
-            csum += [red(rgb), green(rgb), blue(rgb)]
+            csum[1] += red(rgb) # avoid array allocation
+            csum[2] += green(rgb)
+            csum[3] += blue(rgb)
```
kimikage: These are the *left‐off goods* of PR #346.
Modification about `Float32[0, 0, 0]` or `hex(RGB{Float32}(...))` in the outer loop had no effect.

kimikage:
Thank you for your review!
This is not a bug fix but a pure enhancement, so I thought there was no need to hurry. I was rather worried about putting a burden on you.
I hope a little more developers will participate in these discussions.

Issue #349, opened 2019-09-11
`MSC()` is strange


kimikage:
`MSC()` have been introduced to realize the colormap function since 3e2dcf10c96898fba2ee41f27fbc2a8023d571b8 .
I think `MSC()` is strange in some respects.

## 1. Name
As `MSC()` is not a constructor but an ordinaryl function, it is to be desired that the name is lowercase to follow the [Style Guide](https://docs.julialang.org/en/v1/manual/style-guide/index.html).
In particular, I think the naming is important because `MSC` is exported.
However, renaming `MSC` to `msc`, `most_saturated_color`, `maximally_saturated_color` or something else is not sufficient to solve the problems. The reasons are as follows.

By the way, although "colorfulness", "chroma" and "saturation" are often used loosely, the term "chroma" is used in the CIELAB and CIELUV color spaces.

## 2. Return value
`MSC(h)` returns `LCHuv` color, but `MSC(h, l)` returns saturation value.
So, if we follow the behavior, they must have different names and especially the latter should be renamed.
Do we really need two different functions?

## 3. Color space
The current `MSC()` calculates in `Luv`(`LCHuv`) color space. I think the behavior is OK, but the name `MSC()` and its arguments are not informative.
The function to get the maximally saturated color in `Lab` color space, might improve `distinguishable_colors()`.
When we add such a variant function or method, we should modify the interface.

## 4. Validity
I wrote the following ugly function using binary search:
```julia
function find_maximum_chroma(c::LCHuv, low, high)
    high-low < 1e-4 && return low

    mid = (low + high) / 2
    lchm = LCHuv(c.l, mid, c.h)
    rgbm = convert(RGB, lchm)
    notclamped = max(red(rgbm), green(rgbm), blue(rgbm)) < 1 &&
                 min(red(rgbm), green(rgbm), blue(rgbm)) > 0
    if notclamped || ≈(lchm, convert(LCHuv, rgbm), atol=1e-4)
        return find_maximum_chroma(c, mid, high)
    else
        return find_maximum_chroma(c, low, mid)
    end
end
```

And then I got some strange results: 
```julia
julia> find_maximum_chroma(LCHuv(90, NaN, 0), 0, 180) # 180 >= the maximum chroma in sRGB
22.20139503479004

julia> MSC(0, 90)
32.79945146698043

julia> convert(RGB, LCHuv(90, 22, 0)) # not saturated
[...]

PR #350, opened 2019-09-14
Fix erroneous `MSC(h)` and improve accuracy of `MSC(h, l)` (#349)

kimikage:
This focuses primarily on bug fixes mentioned in #349, and the Interface improvements are future tasks.
Therefore, the interfaces remain except the additional keyword argument `linear`.


src/algorithms.jl
```diff
@@ -247,15 +239,55 @@ end
 
 # Maximally saturated color for a specific hue and lightness
 # is found by looking for the edge of LCHuv space.
-function MSC(h,l)
-    pmid=MSC(h)
+function MSC(h, l; linear::Bool=false)
```
kimikage: I was debating but decided to set the default value of `linear` to `false`. The reasons are as follows.
- Since the bug fix of `MSC(h)`  (gamma) is a *slightly* breaking change, it is no longer important whether `MSC(h, l)` involves breaking.
- The differences are *perceptually* insignificant for the preset colormaps.

Issue #351, opened 2019-09-15
[RFC]  Whether RGB <-> XYZ conversion should be more faster or be kept simple


kimikage:
This is not an issue report. And, I have not decided a vision of what it should be.

I was interested in the following comment in another issue.
> But here it doesn't matter because most of the time is taken up by the `v^(1/2.4)` operation in `srgb_compand`.

_Originally posted by @timholy in https://github.com/JuliaGraphics/Colors.jl/issues/349#issuecomment-531460698_

And then, I gave it a try eliminating `^` with `Float32` precision. (I bet there's a better way.)

```julia
function srgb_compand(v::Float32)
    v <= 0.0031308f0 && return 12.92v
    x = sqrt(Float64(v))
    if v >= 0.25f0
        return  -0.027068702898623787 + x*(1.3721685639381944 + x*(
                -0.8182691289384268   + x*(1.0130285191962123 + x*(
                -0.9480751093798352   + x*(0.5864252237648636 + x*(
                -0.2120914317979767   + x*(0.03388206630456692)))))))
    elseif v >= 0.015625f0
        return  -0.04681200154477308  + x*(1.7595676106892333 + x*(
                -4.800335523711974    + x*(28.963281530269672 + x*(
                -144.84684738192738   + x*(551.4601017130509  + x*(
                -1559.1686261466277   + x*(3207.764496252879  + x*(
                -4656.012229629203    + x*(4512.997609562684  + x*(
                -2619.2703123846018   + x*(687.9857377729561)))))))))))
    else
        return  -0.050769586670801364 + x*(2.004280478094793  + x*(
                -11.749002783289283   + x*(146.84612431888658 + x*(
                -1443.135076339118    + x*(9983.922783293772  + x*(
                -45370.79700636592    + x*(121264.00163436493 + x*(
                -144305.83989718332))))))))
    end
end
```
Since the sRGB matrix transformation is (substantially) always done [in `Float64`](https://github.com/JuliaGraphics/Colors.jl/blob/e7f4723e3da81803e8dd392513d11917117d0d21/src/conversions.jl#L140-L147), I changed the type of the matrix elements to `Float32`. (Formally, it may be better to keep them in `Float64` for `XYZ{Float64}`.)

Then, it certainly worked as I expected.

```julia
julia> versioninfo()
Julia Version 1.2.0
Commit c6da87ff4b (2019-08-20 00:03 UTC)
Platform Info:
[...]

PR #352, opened 2019-09-17
Speed up RGB <--> XYZ conversions

kimikage:
This improves `srgb_compand` and `invert_srgb_compand`(renamed rgb->srgb) with almost no loss in accuracy. (#351)

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -73,8 +73,10 @@ end
 correct_gamut(c::CV) where {CV<:AbstractRGB} = CV(clamp01(red(c)), clamp01(green(c)), clamp01(blue(c)))
 clamp01(v::T) where {T<:Fractional} = ifelse(v < zero(T), zero(T), ifelse(v > one(T), one(T), v))
 
-function srgb_compand(v)
-    v <= 0.0031308 ? 12.92v : 1.055v^(1/2.4) - 0.055
+function srgb_compand(v::Fractional)
+    # the following is an optimization technique for `1.055v^(1/2.4) - 0.055`.
+    # x^y ≈ exp(y*log(x)) ≈ exp2(y*log2(y)); the middle form is faster
+    v <= 0.0031308 ? 12.92v : 1.055 * exp(1/2.4 * log(v)) - 0.055
 end
 
 cnvt(::Type{CV}, c::AbstractRGB) where {CV<:AbstractRGB} = CV(red(c), green(c), blue(c))
@@ -166,9 +168,6 @@ end
 cnvt(::Type{CV}, c::LCHab) where {CV<:AbstractRGB}  = cnvt(CV, convert(Lab{eltype(c)}, c))
 cnvt(::Type{CV}, c::LCHuv) where {CV<:AbstractRGB}  = cnvt(CV, convert(Luv{eltype(c)}, c))
 cnvt(::Type{CV}, c::Color3) where {CV<:AbstractRGB}    = cnvt(CV, convert(XYZ{eltype(c)}, c))
-
-cnvt(::Type{CV}, c::RGB24) where {CV<:AbstractRGB{N0f8}} = CV(N0f8((c.color&0x00ff0000)>>>16,0), N0f8((c.color&0x0000ff00)>>>8,0), N0f8(c.color&0x000000ff,0))
-cnvt(::Type{CV}, c::RGB24) where {CV<:AbstractRGB} = CV(((c.color&0x00ff0000)>>>16)/255, (((c.color&0x0000ff00))>>>8)/255, (c.color&0x000000ff)/255)
 
 function cnvt(::Type{CV}, c::AbstractGray) where CV<:AbstractRGB
     g = convert(eltype(CV), gray(c))
@@ -271,12 +270,22 @@ cnvt(::Type{HSI{T}}, c::Color3) where {T} = cnvt(HSI{T}, convert(RGB{T}, c))
 # Everything to XYZ
 # -----------------
 
-function invert_rgb_compand(v)
-    v <= 0.04045 ? v/12.92 : ((v+0.055) /1.055)^2.4
+function invert_srgb_compand(v::Fractional)
+    v <= 0.04045 && return v/12.92
+    # the following is an optimization technique for `((v+0.055) /1.055)^2.4`.
+    # see also: srgb_compand(v::Fractional)
+    x = (v + 0.055) / 1.055
+    return x^2 * exp(0.4 * log(x)) # 2.4 == 2 + 0.4
+end
+
+const invert_srgb_compand_n0f8 = [invert_srgb_compand(v/255) for v = 0:255] # LUT
+
+function invert_srgb_compand(v::N0f8)
+    invert_srgb_compand_n0f8[reinterpret(UInt8, v) + 1]
 end
 
 function cnvt(::Type{XYZ{T}}, c::AbstractRGB) where T
-    r, g, b = invert_rgb_compand(red(c)), invert_rgb_compand(green(c)), invert_rgb_compand(blue(c))
+    r, g, b = invert_srgb_compand(red(c)), invert_srgb_compand(green(c)), invert_srgb_compand(blue(c))
     XYZ{T}(0.4124564*r + 0.3575761*g + 0.1804375*b,
            0.2126729*r + 0.7151522*g + 0.0721750*b,
            0.0193339*r + 0.1191920*g + 0.9503041*b)
@@ -385,8 +394,6 @@ end
 
 cnvt(::Type{XYZ{T}}, c::YIQ) where {T} = cnvt(XYZ{T}, convert(RGB{T}, c))
 cnvt(::Type{XYZ{T}}, c::YCbCr) where {T} = cnvt(XYZ{T}, convert(RGB{T}, c))
-
-cnvt(::Type{XYZ{T}}, c::RGB24) where {T} = cnvt(XYZ{T}, convert(RGB{T}, c))
 
 # Everything to xyY
 # -----------------
--- a/test/conversion.jl
+++ b/test/conversion.jl
@@ -43,6 +43,18 @@ using ColorTypes: eltype_default
     @test hex(RGB(1,0.5,0)) == "FF8000"
     @test hex(RGBA(1,0.5,0,0.25)) == "40FF8000"
 
+    # srgb_compand / invert_srgb_compand
+    @test Colors.srgb_compand(0.5) ≈ 0.7353569830524494 atol=eps()
+    @test Colors.invert_srgb_compand(0.7353569830524494) ≈ 0.5 atol=eps()
+    # issue #351
+    l_pow_x_y() = for i=1:1000; (i/1000)^(1/2.4) end
+    l_exp_y_log_x() = for i=1:1000; exp(1/2.4*log(i/1000)) end
+    l_pow_x_y(); t_pow_x_y = @elapsed l_pow_x_y()
+    l_exp_y_log_x(); t_exp_y_log_x = @elapsed l_exp_y_log_x()
+    if t_exp_y_log_x > t_pow_x_y
+        @warn "Optimization in `[invert_]srgb_compand()` may have the opposite effect."
+    end
+
     fractional_types = (RGB, BGR, RGB1, RGB4)  # types that support Fractional
 
     red24 = reinterpret(RGB24, 0x00ff0000)


kimikage:
# Benchmark
```julia
julia> versioninfo()
Julia Version 1.2.0
Commit c6da87ff4b (2019-08-20 00:03 UTC)
Platform Info:
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```
## before (be3e641d2193df646f1696cdce3d33d8dc62c7ee)
```julia
julia> @benchmark convert(RGB{Float64}, x) setup=(x=XYZ{Float64}(rand(),rand(),rand()))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     76.185 ns (0.00% GC)
  median time:      163.042 ns (0.00% GC)
  mean time:        170.986 ns (0.00% GC)
  maximum time:     435.670 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     970

julia> @benchmark convert(XYZ{Float64}, x) setup=(x=RGB{Float64}(rand(),rand(),rand()))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     5.382 ns (0.00% GC)
  median time:      202.278 ns (0.00% GC)
  mean time:        209.065 ns (0.00% GC)
  maximum time:     391.925 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     966

julia> @benchmark convert(XYZ{Float64}, x) setup=(x=RGB24(rand(),rand(),rand()))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     8.411 ns (0.00% GC)
  median time:      207.477 ns (0.00% GC)
  mean time:        214.632 ns (0.00% GC)
  maximum time:     480.270 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     963
```
## after
```julia
julia> @benchmark convert(RGB{Float64}, x) setup=(x=XYZ{Float64}(rand(),rand(),rand()))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     20.059 ns (0.00% GC)
  median time:      47.944 ns (0.00% GC)
  mean time:        50.982 ns (0.00% GC)
  maximum time:     145.436 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     997

julia> @benchmark convert(XYZ{Float64}, x) setup=(x=RGB{Float64}(rand(),rand(),rand()))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     5.416 ns (0.00% GC)
  median time:      54.764 ns (0.00% GC)
  mean time:        56.797 ns (0.00% GC)
  maximum time:     159.478 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     997

julia> @benchmark convert(XYZ{Float64}, x) setup=(x=RGB24(rand(),rand(),rand()))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     6.899 ns (0.00% GC)
  median time:      7.600 ns (0.00% GC)
  mean time:        7.897 ns (0.00% GC)
  maximum time:     46.901 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

src/conversions.jl
```diff
@@ -167,9 +169,6 @@ cnvt(::Type{CV}, c::LCHab) where {CV<:AbstractRGB}  = cnvt(CV, convert(Lab{eltyp
 cnvt(::Type{CV}, c::LCHuv) where {CV<:AbstractRGB}  = cnvt(CV, convert(Luv{eltype(c)}, c))
 cnvt(::Type{CV}, c::Color3) where {CV<:AbstractRGB}    = cnvt(CV, convert(XYZ{eltype(c)}, c))
 
-cnvt(::Type{CV}, c::RGB24) where {CV<:AbstractRGB{N0f8}} = CV(N0f8((c.color&0x00ff0000)>>>16,0), N0f8((c.color&0x0000ff00)>>>8,0), N0f8(c.color&0x000000ff,0))
```
kimikage: I think this is actually equivalent to `cnvt(::Type{CV}, c::AbstractRGB) where {CV<:AbstractRGB} = CV(red(c), green(c), blue(c))`(L82) and is not helpful for resolving ambiguity.
However, I don't have confidence. Is there a systematic and formal technique for verification?

src/conversions.jl
```diff
@@ -167,9 +169,6 @@ cnvt(::Type{CV}, c::LCHab) where {CV<:AbstractRGB}  = cnvt(CV, convert(Lab{eltyp
...
-cnvt(::Type{CV}, c::RGB24) where {CV<:AbstractRGB} = CV(((c.color&0x00ff0000)>>>16)/255, (((c.color&0x0000ff00))>>>8)/255, (c.color&0x000000ff)/255)
```
kimikage: Is there the reason to convert to `Float64` instead of `N08f`?

src/conversions.jl
```diff
@@ -386,8 +395,6 @@ end
 cnvt(::Type{XYZ{T}}, c::YIQ) where {T} = cnvt(XYZ{T}, convert(RGB{T}, c))
 cnvt(::Type{XYZ{T}}, c::YCbCr) where {T} = cnvt(XYZ{T}, convert(RGB{T}, c))
 
-cnvt(::Type{XYZ{T}}, c::RGB24) where {T} = cnvt(XYZ{T}, convert(RGB{T}, c))
```
kimikage: Now `cnvt(::Type{XYZ{T}}, c::AbstractRGB)` welcomes `RGB{N08f}`.

src/conversions.jl
timholy: This might show you a couple of options for checking:
```julia
julia> using ColorTypes, FixedPointNumbers

julia> cnvt1(::Type{CV}, c::RGB24) where {CV<:AbstractRGB{N0f8}} = CV(N0f8((c.color&0x00ff0000)>>>16,0), N0f8((c.color&0x0000ff00)>>>8,0), N0f8(c.color&0x000000ff,0))
cnvt1 (generic function with 1 method)

julia> cnvt2(::Type{CV}, c::RGB24) where {CV<:AbstractRGB} = CV(((c.color&0x00ff0000)>>>16)/255, (((c.color&0x0000ff00))>>>8)/255, (c.color&0x000000ff)/255)
cnvt2 (generic function with 1 method)

julia> x = RGB24(0, 0, 0)
RGB24(0.0N0f8,0.0N0f8,0.0N0f8)

## Method 1: look at the generated code

julia> @code_llvm cnvt1(RGB{N0f8}, x)

;  @ none:1 within `cnvt1'
define { { i8 }, { i8 }, { i8 } } @julia_cnvt1_16624(%jl_value_t addrspace(10)*, { i32 } addrspace(11)* nocapture nonnull readonly dereferenceable(4)) {
top:
; ┌ @ Base.jl:20 within `getproperty'
   %2 = getelementptr inbounds { i32 }, { i32 } addrspace(11)* %1, i64 0, i32 0
; └
; ┌ @ int.jl:293 within `&'
   %3 = load i32, i32 addrspace(11)* %2, align 4
; └
; ┌ @ int.jl:448 within `>>>' @ int.jl:440
   %4 = lshr i32 %3, 16
; └
; ┌ @ /home/tim/.julia/packages/FixedPointNumbers/870tH/src/normed.jl:7 within `Type'
; │┌ @ int.jl:454 within `rem'
    %5 = trunc i32 %4 to i8
; └└
; ┌ @ int.jl:448 within `>>>' @ int.jl:440
   %6 = lshr i32 %3, 8
; └
; ┌ @ /home/tim/.julia/packages/FixedPointNumbers/870tH/src/normed.jl:7 within `Type'
; │┌ @ int.jl:454 within `rem'
    %7 = trunc i32 %6 to i8
    %8 = trunc i32 %3 to i8
; └└
  %.fca.0.0.insert = insertvalue { { i8 }, { i8 }, { i8 } } undef, i8 %5, 0, 0
  %.fca.1.0.insert = insertvalue { { i8 }, { i8 }, { i8 } } %.fca.0.0.insert, i8 %7, 1, 0
  %.fca.2.0.insert = insertvalue { { i8 }, { i8 }, { i8 } } %.fca.1.0.insert, i8 %8, 2, 0
  ret { { i8 }, { i8 }, { i8 } } %.fca.2.0.insert
}

julia> @code_llvm cnvt2(RGB{N0f8}, x)

;  @ none:1 within `cnvt2'
define { { i8 }, { i8 }, { i8 } } @julia_cnvt2_16636(%jl_value_t addrspace(10)*, { i32 } addrspace(11)* nocapture nonnull readonly dereferenceable(4)) {
top:
; ┌ @ Base.jl:20 within `getproperty'
   %2 = getelementptr inbounds { i32 }, { i32 } addrspace(11)* %1, i64 0, i32 0
; └
; ┌ @ int.jl:293 within `&'
   %3 = load i32, i32 addrspace(11)* %2, align 4
; └
; ┌ @ int.jl:448 within `>>>' @ int.jl:440
   %4 = lshr i32 %3, 16
   %5 = and i32 %4, 255
; └
; ┌ @ int.jl:59 within `/'
; │┌ @ float.jl:271 within `float'
; ││┌ @ float.jl:260 within `Type' @ float.jl:66
     %6 = uitofp i32 %5 to double
; │└└
; │ @ int.jl:59 within `/' @ float.jl:401
   %7 = fdiv double %6, 2.550000e+02
; └
; ┌ @ int.jl:448 within `>>>' @ int.jl:440
   %8 = lshr i32 %3, 8
   %9 = and i32 %8, 255
; └
; ┌ @ int.jl:59 within `/'
; │┌ @ float.jl:271 within `float'
; ││┌ @ float.jl:260 within `Type' @ float.jl:66
     %10 = uitofp i32 %9 to double
; │└└
; │ @ int.jl:59 within `/' @ float.jl:401
   %11 = fdiv double %10, 2.550000e+02
; └
; ┌ @ int.jl:293 within `&'
   %12 = and i32 %3, 255
; └
; ┌ @ int.jl:59 within `/'
; │┌ @ float.jl:271 within `float'
; ││┌ @ float.jl:260 within `Type' @ float.jl:66
     %13 = uitofp i32 %12 to double
; │└└
; │ @ int.jl:59 within `/' @ float.jl:401
   %14 = fdiv double %13, 2.550000e+02
; └
  %15 = call { { i8 }, { i8 }, { i8 } } @julia_Type_16637(%jl_value_t addrspace(10)* addrspacecast (%jl_value_t* inttoptr (i64 139813171978912 to %jl_value_t*) to %jl_value_t addrspace(10)*), double %7, double %11, double %14)
  ret { { i8 }, { i8 }, { i8 } } %15
}

## Method 2: benchmark

julia> using BenchmarkTools

julia> @btime cnvt1(RGB{N0f8}, $x)
  0.016 ns (0 allocations: 0 bytes)
RGB{N0f8}(0.0,0.0,0.0)

julia> @btime cnvt2(RGB{N0f8}, $x)
  5.234 ns (0 allocations: 0 bytes)
RGB{N0f8}(0.0,0.0,0.0)
```
While they give the same result, this method is far more efficient. It's a performance optimization.

(The `0.016ns` is not to be taken seriously, anything much smaller than 1ns just means the compiler elided the whole computation during benchmarking.

src/conversions.jl
timholy: It probably would be better to define this one as returning `CV(cvt(RGB{N0f8}, c))`

src/conversions.jl
```diff
@@ -271,12 +270,22 @@ cnvt(::Type{HSI{T}}, c::Color3) where {T} = cnvt(HSI{T}, convert(RGB{T}, c))
...
+    # the following is an optimization technique for `((v+0.055) /1.055)^2.4`.
+    # see also: srgb_compand(v::Fractional)
+    x = (v + 0.055) / 1.055
+    return x^2 * exp(0.4 * log(x)) # 2.4 == 2 + 0.4
+end
+
+const invert_srgb_compand_n0f8 = [invert_srgb_compand(v/255) for v = 0:255] # LUT
+
+function invert_srgb_compand(v::N0f8)
+    invert_srgb_compand_n0f8[v.i + 1]
```
timholy:
```suggestion
    invert_srgb_compand_n0f8[retinterpret(UInt8, v) + 1]
```
I know it's longer, but this uses an "official" interface rather than digging into the struct.

Entirely optional, you can leave it like it is.

timholy: Oh, and pre-calculating this seems like a good idea!

src/conversions.jl
kimikage: Thank you for your kind reply and  I am sorry for misleading you.
I understand that the former (`cnvt1`) is more efficient than the latter (`cnvt2`) and this is the raison d'être for the former.
So, my question is why the latter method is defined. If the latter method is *not defined*, the former method loses the raison d'être because there is the higher level method `cnvt(::Type{CV}, c::AbstractRGB) where {CV<:AbstractRGB} = CV(red(c), green(c), blue(c))`, which generates the *same* code.

timholy: Ah, yes, if it generates the same code then it can be eliminated.

src/conversions.jl
kimikage: When the latter method is changed to `CV(cnvt(RGB{N0f8}, c))`, the latter loses the raison d'être, too. This is the reason why I deleted the two methods. However, I am a little confused about the complicated method definitions.

timholy: It's not a short file and it accumulated over time, over which the "technology" in ColorTypes was improving. It's entirely possible there are methods that were necessary formerly but which are no longer needed. As long as the tests pass and you don't notice any performance regressions, I think you're good to go.

kimikage: I agree.  I'll check the test cases and the performance again.

kimikage:
# Assessment of the elimination of the two methods.
As [Codecov says](https://codecov.io/gh/JuliaGraphics/Colors.jl/src/be3e641d2193df646f1696cdce3d33d8dc62c7ee/src/conversions.jl), the current test-set does not have the test cases which call [the two methods](https://github.com/JuliaGraphics/Colors.jl/blob/be3e641d2193df646f1696cdce3d33d8dc62c7ee/src/conversions.jl#L170-L171).

The chains of conversion which may call the two methods should have at least one of the two type of (sub) chain (except the *direct* conversion).
- upstream `X` → `RGB24` → `AbstractRGB`
- `RGB24` → `AbstractRGB` → downstream `Y`

## Direct Conversion
The *direct* means that it is not intermediate conversion, i.e. it starts with `RGB24` and ends with `AbstractRGB`.
Let me get straight to the point, there is no *direct* conversion which call the two methods. The conversion completes within ColorTypes.jl as the following comment suggests:
https://github.com/JuliaGraphics/Colors.jl/blob/be3e641d2193df646f1696cdce3d33d8dc62c7ee/src/conversions.jl#L33-L35

- `convert(AbstractRGB, RGB24())`
  - of course, this calls [`convert(::Type{C}, c::C) where {C<:Colorant} = c`](https://github.com/JuliaGraphics/ColorTypes.jl/blob/7ef745605d6a0897078712a0a7a156c51295b559/src/conversions.jl#L6).(practically, inlined)
- `convert(AbstractRGB{N0f8}, RGB24())`
  - same as above. (`RGB24<:AbstractRGB{N0f8} == true`)
- `convert(RGB, RGB24())`
  - this calls [`ColorTypes._convert(RGB{N0f8}, RGB, RGB24, RGB())`](https://github.com/JuliaGraphics/ColorTypes.jl/blob/7ef745605d6a0897078712a0a7a156c51295b559/src/conversions.jl#L25).
- `convert(RGB{Float32}, RGB24())`
  - almost the same as above except the first argument (`RGB{Float32}`).

(to be continued)

kimikage:
Since I found that the two methods are unused and have no effect on performance, I will, *paradoxically*, revert the elimination.
My original intention was that the two methods should be eliminated as a result of fixing the latter one as I commented in [[1]](https://github.com/JuliaGraphics/Colors.jl/pull/352#discussion_r325633412), [[2]](https://github.com/JuliaGraphics/Colors.jl/pull/352#discussion_r325411756), and [[3]](https://github.com/JuliaGraphics/Colors.jl/pull/352#discussion_r325642711).
Althoug the intention was based on the premise that they may be called, it was different to the reality. So, eliminating [the two](https://github.com/JuliaGraphics/Colors.jl/blob/be3e641d2193df646f1696cdce3d33d8dc62c7ee/src/conversions.jl#L170-L171) and [this](https://github.com/JuliaGraphics/Colors.jl/blob/be3e641d2193df646f1696cdce3d33d8dc62c7ee/src/conversions.jl#L389) are totally different issues. Thus, the issue should be report separately.

I will modify them and `push -f`. Then, let's merge it!

kimikage:
Oh, I'm sorry, but I forgot `rebase`.

timholy:
No worries, merge commits are fine as far as I'm concerned. I decide on a case-by-case basis whether I want to keep commits separate. For single-commit PRs I try to remember, but I forget frequently.

Thanks for tackling this!

kimikage:
Now I understand what happened or what I actually did. I fixed a typo `retinterpret` after `rebase master` and reverting the elimination.  I made a mistake at that time.
I'll be more careful.

Issue #353, opened 2019-09-18
[RFC] Order of alpha in hex notation of TransparentColor (#AARRGGBB or #RRGGBBAA)


kimikage:
I was just wondering...
In Colors.jl v0.5.0-v0.9.6, `hex(::TransparentColor)` returns the hex notation in the "AARRGGBB" style. 
https://github.com/JuliaGraphics/Colors.jl/blob/be3e641d2193df646f1696cdce3d33d8dc62c7ee/src/utilities.jl#L24-L33
(`TransparentColor` means {x | x∊ `Colorant` ∧ x∉`Color`}.)

As it is a simple representation of `ARGB32`, the "AARRGGBB" is commonly used (e.g. in Android SDK).
However, there is another common style "RRGGBBAA", too. The [CSS Color Module Level 4 (Editor's Draft)](https://drafts.csswg.org/css-color-4/#hex-notation) introduced this style, and multiple browsers have already [supported it](https://caniuse.com/#feat=mdn-css_properties_color_alpha_hexadecimal_notation).

So, what about splitting `hex(::TransparentColor)` into two, i.e. `hex(::AlphaColor)` returning "AARRGGBB" and `hex(::ColorAlpha)` returning "RRGGBBAA"?

However, this is a breaking change and may affect to the color parsing.
The current Colors.jl does not support parsing the 8-digit or 4-digit hex notation.
```julia
julia> hex(colorant"#FF8000demo") # this is parsed as a opaque orange `RGB{N0f8}` without errors.
"FF8000"

julia> hex(colorant"0xFF8000demo") # same as above
"FF8000"

julia> colorant"reddwarf" # this is not parsed as "red", though.
ERROR: LoadError: Unknown color: reddwarf
```
(I think there are some *minor* problems with the color parser. I'll report a new issue and send a PR.)

Since the "#RRGGBB" notation comes from the CSS Level 3, I think "#FF8000DE" should be parsed as a transparent orange (`RGBA{N0f8}`) according to the CSS manner.
On the other hand, since the "0xRRGGBB"  notation comes from the binary representation, I think "0xFF8000DE" should be parsed as a opaque purple (`ARGB{N0f8}`).
Although this may cause confusion and need the docs, the correspondence between `hex()` and `parse()` is clear.

FYI, "rgba()", "hsla()" and "transparent" are parsed into `RGBA{N0f8}` (not `ARGB{N0f8}`). So, this is compatible with my plan.
```julia
[...]

PR #360, opened 2019-11-03
Modify strictness of color parser (#357)

### LINKED ISSUE #357: Color parser is both strict and loose
The following styles are *generally* valid according to the CSS Level 3, but are rejected in Colors.jl.
```julia
julia> parse(RGBA,"rgba(1,10,100,.5)")
ERROR: Unknown color: rgba(1,10,100,.5)

julia> parse(RGBA,"rgba(1,10,100,\t0.5)")
ERROR: Unknown color: rgba(1,10,100,       0.5)
```

Although we do not have to support parsing the all numerical notations, I think it is worth supporting the former because the former is shorter than something like `+5e-1`.

On the other hand, the following styles are accepted without errors or warnings.
```julia
julia> parse(Colorant,"#FF8000demo")
RGB{N0f8}(1.0,0.502,0.0)

julia> parse(RGB,"p ink")
RGB{N0f8}(1.0,0.753,0.796)
```

The former is potentially dangerous since Colors.jl does not support parsing the 8-digit or 4-digit hex notation yet, as mentioned in #353.
The latter is not so dangerous, but the `replace()` is slightly slow.
https://github.com/JuliaGraphics/Colors.jl/blob/d64fb1a6bd88c4374d72574744c27a7539a4a8ca/src/parse.jl#L50
In the cases where the regex matching is used, there is no need to `replace` the input string. In the case of the named colors, `strip()`, which returns `SubString`, may be helpful.

kimikage:
This fixes #357.
- Modify white space handling
- Add error message about 8-digit or 4-digit hex notation
- Add support for some CSS4 notations
- Add tests
- Speed up parsing averagely

### DIFF
--- a/test/parse.jl
+++ b/test/parse.jl
@@ -4,23 +4,46 @@ using FixedPointNumbers
     r8(x) = reinterpret(N0f8, x)
 
     # Color parsing
+    # named-color
     redN0f8 = parse(Colorant, "red")
     @test colorant"red" == redN0f8
     @test isa(redN0f8, RGB{N0f8})
     @test redN0f8 == RGB(1,0,0)
     @test parse(RGB{Float64}, "red") === RGB{Float64}(1,0,0)
     @test isa(parse(HSV, "blue"), HSV)
-    @test parse(Colorant, "rgb(55,217,127)") === RGB{N0f8}(r8(0x37),r8(0xd9),r8(0x7f))
-    @test colorant"rgb(55,217,127)" === RGB{N0f8}(r8(0x37),r8(0xd9),r8(0x7f))
-    @test parse(Colorant, "rgba(55,217,127,0.5)") === RGBA{N0f8}(r8(0x37),r8(0xd9),r8(0x7f),0.5)
-    @test parse(Colorant, "rgb(55,217,127)") === RGB{N0f8}(r8(0x37),r8(0xd9),r8(0x7f))
-    @test parse(Colorant, "rgba(55,217,127,0.5)") === RGBA{N0f8}(r8(0x37),r8(0xd9),r8(0x7f),0.5)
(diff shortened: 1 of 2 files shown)


kimikage:
It is not difficult to implement the 8-digit or 4-digit hex notation, but we should settle the issue #353.

test/parse.jl
```diff
@@ -4,23 +4,46 @@ using FixedPointNumbers
...
+    @test parse(Colorant, "\nSeaGreen ") === RGB{N0f8}(r8(0x2E),r8(0x8B),r8(0x57))
+
+    # hex-color
     @test parse(Colorant, "#D0FF58") === RGB(r8(0xD0),r8(0xFF),r8(0x58))
+    @test parse(Colorant, "0xd0ff58") === RGB(r8(0xD0),r8(0xFF),r8(0x58))
     @test parse(Colorant, "#FB0") === RGB(r8(0xFF),r8(0xBB),r8(0x00))
+    @test_throws ErrorException parse(Colorant, "#FB0A")
+    @test_throws ErrorException parse(Colorant, "#BAD05")
+    @test_throws ErrorException parse(Colorant, "#BAD0007")
+    @test_throws ErrorException parse(Colorant, "#FFBB00AA") # not supported yet
```
timholy: In such cases is `@test_broken` more appropriate?

kimikage: That's right, if you have already specified the *correct* behavior. Throwing the "not supported" error is not a *broken* behavior *now*. (Of course, checking the error message might be *better*.)

timholy: I'm fine with it either way, no reason not to merge.

kimikage:
If this PR causes a problem (e.g. "not supported" error), it is probably a potential bug in the user scripts.
Although this behavior is going to be changed soon, but I will merge this foremost.

kimikage:
Aww, I'm sorry but I mistaken this PR number and the issue number.

PR #362, opened 2019-11-22
Add [compat] section to Project.toml

timholy:
To silence the emails from CompatHelper!


kimikage:
BTW, Julia v1.3.0 was released today.:tada: 
@timholy, will you run the CI tests on v1.2 for the time being?

timholy:
No reason not to start migrating them to 1.3. AFAICT, Appveyor supports `julia = 1` to mean the latest stable release, but I don't think Travis supports that. Would be nice.

PR #371, opened 2019-12-07
[RFC] Add support for parsing 8-digit and 4-digit hex notations

kimikage:
This adds the support for parsing 8-digit and 4-digit hex notations following the manner mentioned in https://github.com/JuliaGraphics/Colors.jl/issues/353#issue-495280410.

Do not merge this PR before the decision of #353 for the consistency.

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -15,15 +15,16 @@ chop1(x) = SubString(x, 1, lastindex(x) - 1) # `chop` is slightly slow
 
 # Parse a number used in the "rgb()" or "hsl()" color.
 function parse_rgb(num::AbstractString)
-    if num[end] == '%'
+    if @inbounds num[end] == '%'
         return N0f8(clamp(parse(Int, chop1(num), base=10) / 100, 0, 1))
     else
-        return reinterpret(N0f8, UInt8(clamp(parse(Int, num, base=10), 0, 255)))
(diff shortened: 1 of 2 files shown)


kimikage:
Please note the following:
```julia
julia> parse(Colorant, "0xFF8800AA") # opaque purple
ARGB{N0f8}(0.533,0.0,0.667,1.0)

julia> parse(RGBA, "0xFF8800AA") # still opaque purple
RGBA{N0f8}(0.533,0.0,0.667,1.0)
```
We have another option, which refers to the first argument of `parse()`. However, I don't think it is good to depend on the context which doesn't appear in the string.

kimikage:
Perhaps there is little difference in terms of the speed on Julia v1.3.0 and v1.0.5. They are slower than v1.2.0 (especially on Windows) for some unclear reason.

I added `@inbounds` because the boundary checks are very disturbing when trying to clarify the cause of stall or the issue https://github.com/JuliaLang/julia/issues/34055, rather than because of the speed. (In other words, the native codes are quite different.)

Since I don't have enough time to investigate and the cause has not been identified, `@inbounds` has actually little benefit for the end users for now.

kimikage:
I merged #390. I also changed `!= nothing` to `!== nothing` and removed some `@inbounds`.

kimikage:
Since PR #387 was merged, we can merge this.

Issue #333, opened 2019-06-06
Dark gray is light


yha:
`colorant"darkgray"` is lighter than `colorant"gray"`, which seems odd. It's also much lighter than I'd expect for a "dark" gray. Other dark variants are also a bit inconsistent: "darkblue" is much darker than "blue", but "darkorange" is only slightly darker than "orange".
I suppose it could be reasonable if the names come from some standard palette, but if they do, it doesn't seem to be mentioned in the docs.

cormullion:
Coincidentally I read this earlier this month

>Well basically it boils down to the a bit of legacy awkwardness. The first versions of Mosaic and Netscape Navigator were X Applications, meaning they used the X Windows System on UNIX which defines a set of normalized color names which the browsers pulled in.
Eventually the World Wide Web Consortium (W3C) moved to standardize and merged in the X11 color set, but there were some clashes with what the specification wanted.
In HTML “gray” was reserved for the 128 triplet (as in 50%) gray while X11 defined “gray” as a shade around 75% gray which meant since it was reserved it became a shade of 50% gray throwing away the X11 color value.
On the other hand, “dark-gray” was left alone as it did not exist in HTML or CSS at the time so it was brought in from the X11 definition where it is defined as 66% shade of gray.
As a result, the combined CSS color list that prevails on the web today renders “dark-gray” as a significantly lighter shade than “gray” because “dark-gray” came from X11 while “gray” came from HTML.

If you can think of a suitable place in the docs to mention it, it sounds like a good PR...!
(https://medium.com/commitlog/why-dark-gray-is-brighter-than-gray-on-the-web-cab5b933ec1a)

Issue #359, opened 2019-10-03
Improvement of named color charts


kimikage:
It has been on my mind for a while, but the order of colors in [the named color charts](http://juliagraphics.github.io/Colors.jl/v0.9/namedcolors/) is perceptually strange. For example, "hotpink4" is darker than "deeppink2".
![pinks](http://juliagraphics.github.io/Colors.jl/v0.9/assets/figures/namedcolorchart-pinks.svg)

Of course, I think it is not significant, [*too, because the charts are only quick-reference/pretty handouts*](https://github.com/JuliaGraphics/Colors.jl/pull/20#issue-46149952).

Aside from that, I have a suggestion. What about embedding inline SVGs in the HTML document? Although the letters in the current charts are outlined, some browsers allow you to copy and search the embedded SVG texts. Some condensed fonts are available as web fonts.
When we embed the charts, the interactive code technique may be helpful.
https://juliadocs.github.io/Documenter.jl/dev/showcase/#Running-interactive-code-1
Although
> This showcase page is included in Documenter version 0.24.0 and above

AFAIK, the technique is available on the stable version.

---
FYI, the cause of the strange order is incorrect scaling `/255`.
https://github.com/JuliaGraphics/Colors.jl/blob/43d93ffbe855b298fceda03cb2d730b1c223981d/docs/src/assets/figures/generate-named-color-charts.jl#L3-L8
https://github.com/JuliaGraphics/Colors.jl/blob/43d93ffbe855b298fceda03cb2d730b1c223981d/docs/src/assets/figures/generate-named-color-charts.jl#L58-L60

kimikage:
Experimentally I tried embedding the colormaps into document as SVGs using Documenter.jl v0.23.3.

![colormaps](https://user-images.githubusercontent.com/12679384/66189861-abfde600-e6c5-11e9-9b3f-a488f1af5813.png)

"Blues" and "Purples" seem to have changed somewhere from https://github.com/JuliaGraphics/Colors.jl/commit/ebaf8bf9b6d989f4486f53993842b7f6635f6eca to v0.9.6. In order to find such changes, I think it is useful to generate figures dynamically.

However, embedding SVGs is not compatible with the table or list layout of Markdown. In the case of named color charts, it does not matter, though. Of course, it is also OK to embed dynamically generated figures with [`![](path)`](https://juliadocs.github.io/Documenter.jl/v0.23/man/syntax/#@eval-block-1).

[...]

PR #374, opened 2019-12-10
Use inline SVG figures in the documents

kimikage:
cf. #359


cormullion:
Looks terrific! The only thing I'd change (if I was doing it) would be to use spaces instead of commas between the RGB values (partly because I'm a "space after a comma" person (`(r, g, b > (r,g,b)`), which can be frustrating when Julia doesn't do it properly either :)) and because I think it might be easier to read.

Screenshot 2019-12-10 at 09 57 39

I think I used a monospaced font in the old version - these always give good clear spaces.
 
Good to go otherwise!

kimikage:
I agree with you in terms of readability. The reason I didn't use spaces is the ease of mouse (or touch) selection for SVG texts. On Google Chrome, you can select the RGB values as "one" word by double-clicking. However, other browsers do not have such an advantage.

If there is no other opinion, I would like to prioritize the readability.

In addition, I'm going to make some minor fixes regarding the SVG size (or scaling).

cormullion:
I don’t have very strong opinions, so would be happy with any readable solution. I  don’t use Chrome, and  never type RGB values if there’s a simple string name to be used instead. 😂

I’d delete everything unused. Is there a good reason not to?

docs/src/namedcolors.md
```diff
@@ -21,40 +21,86 @@ RGB{N0f8}(0.804,0.361,0.361)
...
+```@example chart
+NamedColorCharts.ColorChartSVG("grays") # hide
+```
+
+!!! info
+    Colors.jl supports the CSS/SVG named colors and the X11 named colors. The
+    CSS/SVG named colors come from the 16 colors defined in HTML3.2 and the X11
+    named colors. There are some unnatural definitions due to the different
+    origins. For example, "LightGray" is lighter than "Gray", but "DarkGray" is
+    also lighter than "Gray".
```
kimikage: @yha, is it okay to close #333?

yha: Yeah, seems fine to me

kimikage:
If you don't like something in docs, please feel free to rewrite it.

PR #376, opened 2019-12-15
Export `CIE2006_2_CMF` and `CIE2006_10_CMF` (Fixes #375)

### LINKED ISSUE #375: Export of `CIE2006_2_CMF` and `CIE2006_10_CMF`
CIE2006 XYZ CMFs for 2° and 10° observers were added in https://github.com/JuliaGraphics/Colors.jl/commit/4ef1406aff158cda0b1abea6deb0456afaf66cc4, but they have not been exported.
https://github.com/JuliaGraphics/Colors.jl/blob/c9f4349bd47efe4d37749bbb1e8ef1a8ad89a743/src/Colors.jl#L23

Moreover, there is a typo.
https://github.com/JuliaGraphics/Colors.jl/blob/c9f4349bd47efe4d37749bbb1e8ef1a8ad89a743/src/colormatch.jl#L1178-L1182

So, we should add the following minimal tests:
```julia
@testset "Colormatch" begin
    @test colormatch(500) === colormatch(CIE1931_CMF, 500)

    cmfs = (CIE1931_CMF, CIE1964_CMF, CIE1931J_CMF, CIE1931JV_CMF, CIE2006_2_CMF, CIE2006_10_CMF)
    @testset "$cmf" for cmf in cmfs
        @test colormatch(cmf, 350.0) === XYZ(0.0, 0.0, 0.0)
        @test colormatch(cmf, 850.0) === XYZ(0.0, 0.0, 0.0)
        xyz450 = colormatch(cmf, 450.0)
        @test xyz450.y < xyz450.x < xyz450.z
        xyz550 = colormatch(cmf, 550.0)
        @test xyz550.z < xyz550.x < xyz550.y
        xyz600 = colormatch(cmf, 600.0)
        @test xyz600.z < xyz600.y < xyz600.x
        xyz551 = colormatch(cmf, 551.0)
        @test colormatch(cmf, 550.5) ≈ (xyz550 + colormatch(cmf, 551)) / 2
    end
end
```

kimikage:
This fixes #375.

### DIFF
--- a/src/Colors.jl
+++ b/src/Colors.jl
@@ -21,7 +21,7 @@ export weighted_color_mean,
        distinguishable_colors, whitebalance,
        colordiff, DE_2000, DE_94, DE_JPC79, DE_CMC, DE_BFD, DE_AB, DE_DIN99, DE_DIN99d, DE_DIN99o,
        MSC, sequential_palette, diverging_palette, colormap,
-       colormatch, CIE1931_CMF, CIE1964_CMF, CIE1931J_CMF, CIE1931JV_CMF
+       colormatch, CIE1931_CMF, CIE1964_CMF, CIE1931J_CMF, CIE1931JV_CMF, CIE2006_2_CMF, CIE2006_10_CMF
 
 # Early utilities
 include("utilities.jl")
--- a/src/colormatch.jl
+++ b/src/colormatch.jl
@@ -15,16 +15,24 @@ Evaluate the CIE standard observer color match function.
 
 # Arguments
 
-- matchingfunction (optional): a type used to specify the matching function. Choices include:
-
-- - `CIE1931_CMF` (the default, the CIE 1931 2° matching function)
-- - `CIE1964_CMF` (the CIE 1964 10° color matching function)
-- - `CIE1931J_CMF` (Judd adjustment to `CIE1931_CMF`)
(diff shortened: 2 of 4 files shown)


timholy:
I certainly have no objections to merging this, but I don't know enough about these matters to comment usefully.

kimikage:
In fact, I don't know whether it is worth exporting these two CMFs. Perhaps when dealing with CMFs other than `CIE1931_CMF`, we should carefully consider the `XYZ` <--> `LMS` conversions, in other words, *the chromatic adaptation*.

I simply want to fix the typo, and I am also concerned that there are no tests for `colormatch`.

However, my additional tests are much looser than the impression we get from the coverage figures. Because the CMFs are defined experimentally rather than theoretically, it is difficult to write tests. I think it is one measure to put the charts like [this](https://commons.wikimedia.org/wiki/File:CIE_1931_XYZ_Color_Matching_Functions.svg#/media/File:CIE_1931_XYZ_Color_Matching_Functions.svg) (dynamically) into the docs, but few people can discuss the differences between the charts.

kimikage:
I fixed a typo and add the note about the standardization.

kimikage:
(I fixed a typo.)

Perhaps other packages do not *directly* depend on the definitions of the CFMs, so I don't think there are compatibility issues with exporting the two CFMs. For the same reason, we don't have to export them, but the export makes the tests and docs simpler.

So, I will do the export.

PR #380, opened 2019-12-15
Use RGB accessor functions Gray conversion

### LINKED ISSUE #377: Problems with `RGB24`->`Gray` and `RGB`->`Gray24` conversions
An error occurs when converting an `RGB24` color to `Gray`. 
```julia
julia> convert(Gray, RGB24(1,0,0))
ERROR: type RGB24 has no field r
```
https://github.com/JuliaGraphics/Colors.jl/blob/d64fb1a6bd88c4374d72574744c27a7539a4a8ca/src/conversions.jl#L783-L787

This may not be an important issue, but this is a good example to show the importance of the unveiling.
I think we should avoid the heavy use of multiple dispatch for the *public* `convert()` .

*BTW, I think there is no need for the `min`.*

There is a similar problem with conversion to `Gray24` because `Gray24` is not a subtype of `Gray`, but a subtype of `AbstractGray`.
```julia
julia> convert(Gray24, RGB())
ERROR: StackOverflowError:
```

_Originally posted by @kimikage in https://github.com/JuliaGraphics/Colors.jl/issues/354#issuecomment-534923006_

timholy:
This also provides methods for Gray24 conversion.  Fixes #377.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -779,11 +779,17 @@ end
 const unsafe_trunc = Base.unsafe_trunc
 
 convert(::Type{Gray{T}}, x::Gray{T}) where {T} = x
-convert(::Type{Gray{T}}, x::Gray) where {T} = Gray{T}(gray(x))
-function convert(::Type{Gray{T}}, x::AbstractRGB{T}) where {T<:Normed}
-    TU = FixedPointNumbers.rawtype(T)
-    val = min(typemax(TU), 0.299f0*reinterpret(x.r) + 0.587f0*reinterpret(x.g) + 0.114f0*reinterpret(x.b))
-    return Gray{T}(T(round(TU, val), 0))
+convert(::Type{Gray24}, x::Gray24) = x
+
+convert(::Type{G}, x::AbstractGray) where {G<:AbstractGray} = G(gray(x))
+
+function convert(::Type{G}, x::AbstractRGB{T}) where {G<:AbstractGray,T<:Normed}
+    TU, Tf = FixedPointNumbers.rawtype(T), floattype(T)
+    val = min(typemax(TU), Tf(0.299)*reinterpret(red(x)) + Tf(0.587)*reinterpret(green(x)) + Tf(0.114)*reinterpret(blue(x)))
+    return G(reinterpret(T, round(TU, val)))
 end
-convert(::Type{Gray{T}}, x::AbstractRGB) where {T} = convert(Gray{T}, 0.299f0*x.r + 0.587f0*x.g + 0.114f0*x.b)
-convert(::Type{Gray{T}}, x::Color) where {T} = convert(Gray{T}, convert(RGB{T}, x))
+convert(::Type{G}, x::AbstractRGB) where {G<:AbstractGray} =
+    G(0.299f0*red(x) + 0.587f0*green(x) + 0.114f0*blue(x))
+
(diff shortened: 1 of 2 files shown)


src/conversions.jl
```diff
@@ -781,9 +781,12 @@ const unsafe_trunc = Base.unsafe_trunc
...
-    val = min(typemax(TU), 0.299f0*reinterpret(x.r) + 0.587f0*reinterpret(x.g) + 0.114f0*reinterpret(x.b))
+    TU, Tf = FixedPointNumbers.rawtype(T), floattype(T)
+    val = min(typemax(TU), Tf(0.299)*reinterpret(red(x)) + Tf(0.587)*reinterpret(green(x)) + Tf(0.114)*reinterpret(blue(x)))
     return Gray{T}(T(round(TU, val), 0))
 end
 convert(::Type{Gray{T}}, x::AbstractRGB) where {T} = convert(Gray{T}, 0.299f0*x.r + 0.587f0*x.g + 0.114f0*x.b)
 convert(::Type{Gray{T}}, x::Color) where {T} = convert(Gray{T}, convert(RGB{T}, x))
+
+convert(::Type{Gray24}, c::Color) = Gray24(convert(Gray{N0f8}, c))
+convert(::Type{Gray24}, c::Gray{N0f8}) = Gray24(gray(c))
```
kimikage: This is a *right* bug fix, but is there any practical benefit to this specialization? We have `AbstractGray`.

src/conversions.jl
```diff
@@ -781,9 +781,12 @@ const unsafe_trunc = Base.unsafe_trunc
 convert(::Type{Gray{T}}, x::Gray{T}) where {T} = x
 convert(::Type{Gray{T}}, x::Gray) where {T} = Gray{T}(gray(x))
 function convert(::Type{Gray{T}}, x::AbstractRGB{T}) where {T<:Normed}
-    TU = FixedPointNumbers.rawtype(T)
-    val = min(typemax(TU), 0.299f0*reinterpret(x.r) + 0.587f0*reinterpret(x.g) + 0.114f0*reinterpret(x.b))
+    TU, Tf = FixedPointNumbers.rawtype(T), floattype(T)
+    val = min(typemax(TU), Tf(0.299)*reinterpret(red(x)) + Tf(0.587)*reinterpret(green(x)) + Tf(0.114)*reinterpret(blue(x)))
     return Gray{T}(T(round(TU, val), 0))
```
kimikage: Although we don't have to do it here, what about using `reinterpret(T, round(TU, val))`?

kimikage: > *BTW, I think there is no need for the `min`.*

```julia
julia> round(UInt8, 0.299f0 * 0xFF + 0.587f0 * 0xFF + 0.114f0 * 0xFF)
0xff

julia> round(UInt16, 0.299f0 * 0xFFFF + 0.587f0 * 0xFFFF + 0.114f0 * 0xFFFF)
0xffff

julia> round(UInt32, 0.299 * 0xFFFFFFFF + 0.587 * 0xFFFFFFFF + 0.114 * 0xFFFFFFFF)
0xffffffff
```
Of course, this is not essential and just for future reference.

src/conversions.jl
timholy: It was just an easy way to avoid method ambiguities. But I've done a bigger cleanup of the gray conversion methods now.

PR #381, opened 2019-12-17
Version 0.10.0

timholy:
There's a ton of unreleased work here, overwhelmingly thanks to @kimikage: https://github.com/JuliaGraphics/Colors.jl/compare/v0.9.6...master. Seems like it may be time for a new release. Is there anything anyone wants to merge? @kimikage, I note you have some other nice stuff almost ready, so I'll let you make the call.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Colors"
 uuid = "5ae59095-9a9b-59fe-a467-6f913c188581"
-version = "0.9.6"
+version = "0.10.0"
 
 [deps]
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
@@ -12,7 +12,7 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 [compat]
 julia = "1"
 ColorTypes = "0.7, 0.8"
-FixedPointNumbers = "0.6"
+FixedPointNumbers = "0.6, 0.7"
 Reexport = "0.2"
 
 [extras]


kimikage:
I’m good to go. :rocket:

However, if we change the behavior of `hex(c::ColorAlpha)` in v1.0, it might be a good idea to introduce https://github.com/JuliaGraphics/Colors.jl/issues/353#issuecomment-562079254 with `@warn`:
```julia
function hex(c::ColorAlpha)
    @warn("""The alpha position for `$(typeof(c))` (`<:ColorAlpha`) will soon be changed.
             You can get the alpha-first style string by `hex(c, :AARRGGBB)` or `hex(c |> ARGB32)`.""")
    #_hex(HexNotation{RGBA,:upper,8}, c) # breaking change in v1.0
    _hex( HexNotation{ARGB,:upper,8}, c) # backward compatible
end
```

Of course, you don't have to wait just for this because `@warn` is not so effective and I need at least a day to write the document.

Other than `hex`, I have no urgent PR towards v1.0.

timholy:
We could do a deprecation-warning release and leave it for a month or so and then go to 1.0.

This reminds that I needed to do https://github.com/JuliaGraphics/ColorTypes.jl/pull/143

timholy:
BTW, do you know about `@deprecate` and `Base.@deprecate_binding`?

timholy:
Let's get the `hex` deprecation in at the same time that we bump compat for ColorTypes 0.9 (https://github.com/JuliaGraphics/ColorTypes.jl/pull/143). Then we can release Colors 0.11 and make it depend on ColorTypes 0.9 as a minimum version.

kimikage:
> Let's get the hex deprecation in at the same time that we bump compat for ColorTypes 0.9
Understood.

> BTW, do you know about `@deprecate` and `Base.@deprecate_binding`?

I know them, but I don't know whether should we use them or not, because I'm going to keep the 1-arg `hex` itself and I don't know what *actually* the users need.

timholy:
We could use `Base.depwarn`. The problem with `@warn` is it will fire every time the function is called; if one package calls `hex` repeatedly it will be really annoying. Conversely, if we use `Base.depwarn` it will only warn once. Of course, it will drastically slow performance because it collects a backtrace each time as a mechanism to ensure the warn-once behavior.

In either case we need to provide the 2-arg version as a way to circumvent the warning.

PR #383, opened 2019-12-18
Support ColorTypes 0.9

timholy:
One thing I failed to anticipate is that because this reexports names in ColorTypes, I needed to create new deprecated bindings if I didn't want to cause a depwarn every time the package is loaded. It doesn't appear to cause a conflict, though:
```julia
julia> using ColorTypes

julia> RGB1
WARNING: ColorTypes.RGB1 is deprecated, use XRGB instead.
  likely near REPL[2]:1
XRGB

julia> using Colors

julia> RGB1
WARNING: ColorTypes.RGB1 is deprecated, use XRGB instead.
  likely near REPL[4]:1
XRGB
```
and in a fresh session
```julia
julia> using ColorTypes, Colors

julia> RGB1
WARNING: Colors.RGB1 is deprecated, use XRGB instead.
  likely near REPL[2]:1
XRGB
```

Normally I'd say this is unambiguously a patch release (and that's what I've done here), but since there are the "new" bindings I'd be open to making this a minor bump if others think that's more appropriate. (In packages higher up the hierarchy, supporting these new names will be a patch release.)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Colors"
 uuid = "5ae59095-9a9b-59fe-a467-6f913c188581"
-version = "0.10.0"
+version = "0.11.0"
 
 [deps]
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
@@ -10,7 +10,7 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
-ColorTypes = "0.7, 0.8"
+ColorTypes = "0.9"
 FixedPointNumbers = "0.6, 0.7"
 Reexport = "0.2"
 julia = "1"
--- a/docs/src/colorspaces.md
+++ b/docs/src/colorspaces.md
@@ -5,7 +5,7 @@
 
 The colorspaces used by Colors are defined in [ColorTypes](https://github.com/JuliaGraphics/ColorTypes.jl). Briefly, the defined spaces are:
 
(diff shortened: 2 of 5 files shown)


timholy:
I'm going to merge this faster than I otherwise would in case it helps with https://github.com/JuliaGraphics/ColorTypes.jl/pull/143#issuecomment-567136645.

timholy:
But for safety I'll change this to a minor bump.

johnnychen94:
By changing the compat from `ColorTypes = "0.7, 0.8"` to `ColorTypes = "0.9"`, it's already a breaking change for users, so it definitely worth a minor version. If we think one step further, any breaking change of this package should also be a breaking change to `ImageCore`, and hence a breaking change to `Images` unless we fix that (e.g., https://github.com/JuliaImages/ImageCore.jl/pull/107)

kimikage:
BTW, dropping v0.8.0 compatibility means that we can use `hue()` and `chroma()`. :smiley:

timholy:
> Since ColorTypes v0.9 only introduces a depwarn, it's not a breaking change to Colors, so for Colors we don't necessarily need to tag a breaking change version. 

I largely agree, and that's why originally it was a minor version bump. However, to be conservative I decided on minor bump for the following reasons:
- ColorTypes is essentially "half a package", and this is the other half. Many packages don't say `using ColorTypes`, and instead just say `using Colors`. `@reexport` does muddy the versioning a bit...
- Depwarns can absolutely kill performance, though, and for some workloads that's effectively breaking.
- Given that I already had one problem report (https://github.com/JuliaGraphics/ColorTypes.jl/pull/143) I wanted to make sure there was a way to opt out.

But even with all that, one could equally well argue that this was the wrong decision and better to go with the minor bump.

I am on the road a lot for the holidays for the next couple of days so I will be unavailable to respond to any issues that crop up.

johnnychen94:
> I am on the road a lot for the holidays for the next couple of days

It's definitely the time to take a break here. 🎉

PR #387, opened 2019-12-22
Add support for specifying the style of `hex`

kimikage:
cf. issue #353

Since `Base.hex` is no longer public, this `hex` does not extend it (i.e. this is `Colors.hex`).
Because of the backward compatibility, `hex(c::ColorAlpha)` returns "AARRGGBB" with a depwarn for now.

The new methods are not fully optimized, so the performance improvement is a future task.

### DIFF
--- a/src/Colors.jl
+++ b/src/Colors.jl
@@ -2,7 +2,6 @@ module Colors
 
 using FixedPointNumbers
 using Reexport
-using Printf
 
 @reexport using ColorTypes
 Base.@deprecate_binding RGB1 XRGB
@@ -13,7 +12,7 @@ AbstractAGray{C<:AbstractGray,T} = AlphaColor{C,T,2}
 AbstractGrayA{C<:AbstractGray,T} = ColorAlpha{C,T,2}
 
 import Base: ==, +, -, *, /
-import Base: convert, eltype, hex, isless, range, show, typemin, typemax
+import Base: convert, eltype, isless, range, show, typemin, typemax
 
 # Additional exports, not exported by ColorTypes
 export weighted_color_mean,
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -8,29 +8,137 @@ function lerp(x, a, b)
     a + (b - a) * max(min(x, one(x)), zero(x))
 end
 
+"""
+    HexNotation{C, A, N}
+
+This is a private type for specifying the style of hex notations. It is not
+recommended to use this type and its derived types in user scripts or other
+packages, since they may change in the future without notice.
+
+# Arguments
+- `C`: a base colorant type.
+- `A`: a symbol (`:upper` or `:lower`) to specify the letter casing.
+- `N`: a total number of digits.
+"""
+abstract type HexNotation{C, A, N} end
+abstract type HexAuto <: HexNotation{Colorant,:upper,0} end
+abstract type HexShort{A} <: HexNotation{Colorant,A,0} end
 
 """
+    hex(c::Colorant)
+    hex(c::Colorant, style::Symbol)
+
+Convert a color to a hexadecimal string, optionally specifying its style.
+
+# Arguments
+- `c`: a target color.
+- `style`: a symbol to specify the hexadecimal notation. Spesifying the
+  uppercase symbols means the return values are in uppercase. The following
+  symbols are available:
+  - `:AUTO`: notation automatically selected according to the type of `c`
+  - `:RRGGBB`/`:rrggbb`: 6-digit opaque notation
+  - `:AARRGGBB`/`:aarrggbb`: 8-digit notation with alpha at the head
+  - `:RRGGBBAA`/`:rrggbbaa`: 8-digit notation with alpha at the tail
+  - `:RGB`/`:rgb`/`:ARGB`/`:argb`/`:RGBA`/`:rgba`: 3-digit or 4-digit noatation
+  - `:S`/`:s`: short notation if available
+
+# Examples
+```jldoctest; setup = :(using Colors)
+julia> hex(RGB(1,0.5,0))
+"FF8000"
+
+julia> hex(ARGB(1,0.5,0,0.25))
+"40FF8000"
+
+julia> hex(HSV(30,1.0,1.0), :AARRGGBB)
+"FFFF8000"
+
+julia> hex(ARGB(1,0.533,0,0.267), :rrggbbaa)
+"ff880044"
+
+julia> hex(ARGB(1,0.533,0,0.267), :rgba)
+"f804"
+
+julia> hex(ARGB(1,0.533,0,0.267), :S)
+"4F80"
+```
+
+!!! compat
+    For backward compatibility, `hex(c::ColorAlpha)` currently returns an
+    "AARRGGBB" style string. This is inconsistent with `hex(c, :AUTO)` returning
+    an "RRGGBBAA" style string. The alpha position for `ColorAlpha` will soon be
+    changed to the tail.
+"""
+hex(c::Colorant) = _hex(HexAuto, c) # there is no need to search the dictionary
+hex(c::Colorant, style::Symbol) = _hex(get(_hex_styles, style, HexAuto), c)
+
+function Base.hex(c::Colorant)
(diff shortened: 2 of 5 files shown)


src/Colors.jl
```diff
@@ -13,7 +12,7 @@ AbstractAGray{C<:AbstractGray,T} = AlphaColor{C,T,2}
 AbstractGrayA{C<:AbstractGray,T} = ColorAlpha{C,T,2}
 
 import Base: ==, +, -, *, /
-import Base: convert, eltype, hex, isless, range, show, typemin, typemax
+import Base: convert, eltype, isless, range, show, typemin, typemax
```
johnnychen94: Not exporting the name is one of the breaking changes in this PR, so it's better to have some deprecations on this as well 😢

kimikage: Strictly speaking, your point is right, but scripts which use `hex` will eventually need to be recompiled.
I'll try to see if `@deprecate_binding` works.

kimikage: I added a depwarn.

src/utilities.jl
```diff
@@ -8,29 +8,132 @@ function lerp(x, a, b)
...
+    :RGB => HexNotation{RGB,:upper,3}, :rgb => HexNotation{RGB,:lower,3},
+    :ARGB => HexNotation{ARGB,:upper,4}, :argb => HexNotation{ARGB,:lower,4},
+    :RGBA => HexNotation{RGBA,:upper,4}, :rgba => HexNotation{RGBA,:lower,4},
+    :RRGGBB => HexNotation{RGB,:upper,6}, :rrggbb => HexNotation{RGB,:lower,6},
+    :AARRGGBB => HexNotation{ARGB,:upper,8}, :aarrggbb => HexNotation{ARGB,:lower,8},
+    :RRGGBBAA => HexNotation{RGBA,:upper,8}, :rrggbbaa => HexNotation{RGBA,:lower,8},
+)
+@inline function _hexstring(::Type{T}, u::U, itr) where {C, T <: HexNotation{C,:upper}, U <: Unsigned}
+    s = UInt8(8sizeof(u) - 4)
+    @inbounds String([b"0123456789ABCDEF"[u << i >> s + 1] for i in itr])
```
timholy: Optional: clarify the operator precedence with parentheses?

src/utilities.jl
```diff
@@ -8,29 +8,132 @@ function lerp(x, a, b)
...
+    s = u == (u & 0x0F0F0F0F) * 0x11
+    c isa AlphaColor && return _hex(HexNotation{ARGB, A, s ? 4 : 8}, u)
+    c isa ColorAlpha && return _hex(HexNotation{RGBA, A, s ? 4 : 8}, u)
+    _hex(HexNotation{RGB, A, s ? 3 : 6}, u)
+end
+
+function _hex(t::Type{T}, u::UInt32) where {C <:Union{RGB, ARGB, RGBA}, A, T <: HexNotation{C,A}}
+    u64 = UInt64(u)
+    # TODO: use SIMD `move` with zero extension (e.g. vpmovzxbw)
+    unpacked = ((u64 & 0xFF00FF00)<<24) | (u64 & 0x00FF00FF) # 0x00AA00GG00RR00BB
```
timholy: I've wondered if we should have an `ARGB64` that performs this transformation. In principle it's nice for SIMD-like operations (e.g., bilinear interpolation), but packing the values seems to often be more costly than the benefit.

src/utilities.jl
```diff
@@ -8,29 +8,132 @@ function lerp(x, a, b)
...
+
+_hex(t::Type, c::Colorant) = _hex(t, reinterpret(UInt32, ARGB32(c)))
+
+_hex(::Type{HexAuto}, c::Color) = _hex(HexNotation{RGB,:upper,6}, c)
+_hex(::Type{HexAuto}, c::AlphaColor) = _hex(HexNotation{ARGB,:upper,8}, c)
+_hex(::Type{HexAuto}, c::ColorAlpha) = _hex(HexNotation{RGBA,:upper,8}, c)
+
+function _hex(::Type{HexShort{A}}, c::Colorant) where A
+    u = reinterpret(UInt32, ARGB32(c))
+    s = u == (u & 0x0F0F0F0F) * 0x11
```
timholy: elegant

src/utilities.jl
```diff
@@ -8,29 +8,132 @@ function lerp(x, a, b)
...
+    c isa AlphaColor && return _hex(HexNotation{ARGB, A, s ? 4 : 8}, u)
+    c isa ColorAlpha && return _hex(HexNotation{RGBA, A, s ? 4 : 8}, u)
+    _hex(HexNotation{RGB, A, s ? 3 : 6}, u)
+end
+
+function _hex(t::Type{T}, u::UInt32) where {C <:Union{RGB, ARGB, RGBA}, A, T <: HexNotation{C,A}}
+    u64 = UInt64(u)
+    # TODO: use SIMD `move` with zero extension (e.g. vpmovzxbw)
+    unpacked = ((u64 & 0xFF00FF00)<<24) | (u64 & 0x00FF00FF) # 0x00AA00GG00RR00BB
+    q = muladd(unpacked, 0xF,  0x0087_0087_0087_0087)
```
timholy: Not sure I follow from here on, maybe add comments?

kimikage: This is a simple division.:smile:
The SIMD division is not so slow, but the compiler did not vectorize (pack/unpack) well. :sob:
```julia
julia> 256 / 17 # slightly greater than 15
15.058823529411764

julia> all(x -> round(x / 17) == ((x + 9) * 15) >> 8, 0:255)
true

julia> all(x -> round(x / 17) == (x * 15 + 135) >> 8, 0:255)
true
```
I'll add comments that this method is for 3-digit or 4-digit notations and its calculation is "1/17".

src/utilities.jl
kimikage: This is just a SIMD-like operation. Each element requires at least 4-bit padding. (see below)

timholy: I understand, I just meant maybe we want `unpacked = ARGB64(c::Colorant).color`.

kimikage: I see. However, `ARGB64(ARGB32(0x11223344))` should return `0x1111222233334444`, so we will often need an extra mask.:smile:

kimikage:
I added comments.

src/utilities.jl
timholy: I was envisioning ARGB64 as 4 squashed `N8f8`s, so it would return `0x0011002200330044`.

But of course for some computations (e.g., interpolation) we'd need to have a defined mask for stripping the low-order bits from the next color channel.
```julia
function bilinear_interpolate(img::AbstractMatrix{T}, y, x) where T<:AbstractRGB{N0f8}
    xf, yf = floor(x), floor(y)
    Δx, Δy = x-xf, y-yf
    # Get the 4 corners of the array data 
    xl, yl = unsafe_trunc(Int, xf), unsafe_trunc(Int, yf)
    xu, yu = iszero(Δx) ? xl : xl + 1, iszero(Δy) ? yl : yl + 1
    ll, ul = argb64(img[yl, xl]), argb64(img[yu, xl])
    lu, uu = argb64(img[yl, xu]), argb64(img[yu, xu])
    # Convert the fractional part of the position into what's effectively a `Q7f8`:
    Δxi, Δyi = unsafe_trunc(UInt16, 256*Δx), unsafe_trunc(UInt16, 256*Δy)
    Δxic, Δyic = 0x0100 - Δxi, 0x0100 - Δyi # the complement
    # interpolate along the first dimension
    cl = ((Δyic * ll + Δyi * ul) >> 0x08) & 0x00ff00ff00ff00ff
    cu = ((Δyic * lu + Δyi * uu) >> 0x08) & 0x00ff00ff00ff00ff
    # interpolate along the second dimension
    cc = ((Δxic * cl + Δxi * cu) >> 0x08) & 0x00ff00ff00ff00ff
```

timholy:
This seems ready to merge.

kimikage:
Since we chose to use patches (i.e. since we would not need to edit the [compat] section), I think we did all we have to do in Colors.jl. Let's step forward.

PR #388, opened 2019-12-22
[RFC] Add support for parsing X11 color names with spaces

kimikage:
I submit this PR for reference (cf. issue #386), but I don't think this change is really desirable. Therefore, I set this as a "Draft PR".

This adds 116 color names.
- 96 traditional names with spaces
- 10 Web color names with/without spaces
- 10 X11 color names with/without spaces

Issue #386, opened 2019-12-19
[RFC] Re-support of original X11 color names


kimikage:
I ended support for parsing some X11 color names which conflict with the SVG/CSS specification (e.g. "deep sky blue") in PR #360. However, some users or packages use them.
cf. https://github.com/GiovineItalia/Gadfly.jl/issues/1368

I don't think it's a good idea to use the *old* synonyms in *modern times* because it goes against the intention of standardization (formalization) of the SVG/CSS color.
Moreover, in the past, it seems that there is a case where the aliases with spaces have been deleted.
https://web.archive.org/web/20070928030522/http://cvsweb.xfree86.org/cvsweb/xc/programs/rgb/rgb.txt.diff?r1=1.2&r2=1.1

However, I would not object to someone adding them to [`color_names`](https://github.com/JuliaGraphics/Colors.jl/blob/9ba8b264232717f0495dd05becf0e404b699c38b/src/names_data.jl#L3).
I'm against the removal (neglect) of *all* spaces. (e.g. "p ink"->"pink")

timholy:
Sorry I failed to notice this had changed when I reviewed #360. The kindest option is to issue a deprecation warning. If you are ambitious you can do that with a 0.10.1 and 0.11.1 release (you'll need to create some branches and use git-cherry-pick). I won't have time to help due to travel, unfortunately.

kimikage:
I'm sorry, but I didn't think this would be a problem. I was wrong.

I think the deprecation warning is useful only when there are alternatives. For the users who want the traditional names with spaces, the names without spaces should not be alternatives.

Unfortunately, I realized that #374, which designed in anticipation of the parser changes (cf. https://github.com/JuliaGraphics/Colors.jl/issues/359#issuecomment-542987779), was not so useful.:sob:

kimikage:
Perhaps we have four choices.
1. Add the deprecation warning to color parser and release v0.10.1 and v0.11.1
2. Assuming PR #360 is a bug, merge PR #388(-like something) and issue v0.10.1 and v0.11.1
3. As a feature extension, merge PR #388(-like something) and release v0.12.0 with other changes (e.g. #387, #371)
4. Do nothing about this and release v0.12.0 with other changes


I prefer to re-support the traditional colors rather than add warnings. However, we need to be desperately aware that when we support somethings, it will be awkward to break them.

johnnychen94:
[...]

PR #390, opened 2019-12-30
Re-support parsing the X11 color names with spaces with a depwarn

kimikage:
This is another solution for the issue #386. (cf. PR #388)

This parser is somewhat loose. For example, "deep skyblue" or "dark golden rod" are acceptable even though they are not the X11 named colors.

### DIFF
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -122,6 +122,20 @@ function _parse_colorant(desc::AbstractString)
         return RGBA{N0f8}(0,0,0,0)
     end
 
+    wo_spaces = replace(ldesc, r"(?<=[^ ]{3}) (?=[^ ]{3})" => "")
+    c = get(color_names, wo_spaces, nothing)
+    if c != nothing
+        camel = replace(titlecase(ldesc), " " => "")
+        Base.depwarn(
+            """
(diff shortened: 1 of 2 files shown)


src/parse.jl
```diff
@@ -122,6 +122,20 @@ function _parse_colorant(desc::AbstractString)
         return RGBA{N0f8}(0,0,0,0)
     end
 
+    wo_spaces = replace(ldesc, r"(?<=[^ ]{3}) (?=[^ ]{3})" => "")
+    c = get(color_names, wo_spaces, nothing)
+    if c != nothing
```
timholy:
```suggestion
    if c !== nothing
```
In some cases the codegen is better. For example, check
```julia
julia> function checkiter1(iter)
           ret = iterate(iter)
           if ret == nothing
               println("nothing")
           else
               println("something")
           end
           nothing
       end

julia> function checkiter2(iter)
           ret = iterate(iter)
           if ret === nothing
               println("nothing")
           else
               println("something")
           end
           nothing
       end
```
with `@code_llvm checkiterX([])`.

kimikage: Understood. I'll fix it in the PR #371.

PR #395, opened 2020-01-05
Work around segfault on Julia 1.0.x

timholy:
Ref https://github.com/JuliaLang/julia/issues/34121


kimikage:
Thank you for your speedy handling. However, we seem to be still playing "whack-a-mole". :hammer:

timholy:
Fixing Julia will take more time than I have right now.

kimikage:
I'm sorry. I should not have said that joke. I mean https://github.com/GiovineItalia/Gadfly.jl/issues/1368#issuecomment-570911101 .

Issue #379, opened 2019-12-15
Inconsistency in handling hues


kimikage:
Unnormalized hues are somewhat useful in the interpolation. In other words, the interpolation across 0° is not possible with only the normalized hues.

However, the manners of input range checking in conversions vary (cf. #378 ), and some conversions cannot handle unnormalized hues correctly.

```julia
using Colors

range(HSV(0,1,1), HSV(360,1,1),length=120)
range(HSV(0,1,1), HSV(-360,1,1),length=120)
range(HSV(0,1,1), HSV(720,1,1),length=120)
range(HSL(0,1,0.5), HSL(720,1,0.5),length=120)
range(HSI(0,1,1/3), HSI(720,1,1/3),length=120)
range(LCHab(70,70,0), LCHab(70,70,720),length=120)
range(LCHuv(70,70,0), LCHuv(70,70,720),length=120)
```

![hue](https://user-images.githubusercontent.com/12679384/70864262-cee52a00-1f93-11ea-893c-74245108a69b.png)

I think that the strictness of input range checking should be more consistent. Apart from that, it might be a good idea to add some utility functions related to the hue, such as `normalize_hue(c)`.

kimikage:
Just out of my curiosity, @johnnychen94,what are your concerns? Use of unnormalized hues? Backward compatibility? Additional costs for the normalizing?

johnnychen94:
Hmmm, I'm not familiar with this topic, but as far as I can think of, normalizing a channel needs `maxvalue` of that channel, do we have some kind of assumption/defaults on it? For Grays and RGBs we can infer the max value from the storage type (e.g., `N0f8` => 255), but since hue is stored as floats, we generally don't have such information, right? If you need to manually pass a `maxvalue` to `normalize_hue`, it might be a just not so convenient util.

kimikage:
I see. Because of the problems which you point out, I focused only on the hue, which should be simply normalized to [0, 360], not to [0, 1].

BTW, I think the color gamut problem must be solved somewhere (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/140). It is probably too early to put the gamut support in Colors.jl v1.0.

johnnychen94:
I'm thinking of such kind of normalization:

```julia
function normalize_hue(x::T; maxvalue, minvalue) where T<:HSV
h_max= gamutmax(T)[1]
[...]

Issue #378, opened 2019-12-15
HSI->RGB conversion may return a value out of [0,1]


kimikage:
FWIW, `HSI`->`RGB` conversion may return a value out of [0,1].
```julia
julia> convert(RGB, HSV(90,1,1))
RGB{Float32}(0.5f0,1.0f0,0.0f0)

julia> convert(RGB, HSL(90,1,1))
RGB{Float32}(1.0f0,1.0f0,1.0f0)

julia> convert(RGB, HSI(90,1,1))
RGB{Float32}(1.0f0,2.0f0,0.0f0)
```
This is not consistent with other conversions to `RGB`. However, the conversions from `HSV` and `HSL` also do not check the input range.

```julia
julia> convert(RGB, HSV(400,2,2))
RGB{Float32}(2.0f0,0.66666603f0,-2.0f0)

julia> convert(RGB, HSL(400,2,2))
RGB{Float32}(0.0f0,1.3333333f0,4.0f0)

julia> convert(RGB, HSI(400,2,2))
RGB{Float32}(5.26083f0,2.73917f0,-2.0f0)
```

IMO, it makes sense to clamp the values which are not numerically meaningful or the values ​​which exceed `typemin`/`typemax`, but otherwise it is unnecessary.
If we consider color gamuts other than sRGB, we will need to reconsider the clipping.

kimikage:
One use case where this problem become tangible is calling `hex`.
```julia
julia> hex(HSI(90,1,1))
ERROR: ArgumentError: element type FixedPointNumbers.Normed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0,
  but the values (1.0f0, 2.0f0, 0.0f0) do not lie within this range.
  See the READMEs for FixedPointNumbers and ColorTypes for more information.
```

I removed the clamping, i.e. stopped using `lerp`, in PR #387, because:
- the clamping is done in `XYZ`/`YIQ`/`YCbCr` --> `RGB` conversion, i.e. `Lab`, `Luv` and so on also do not require the extra clamping
- the normal `HSV` and `HSL` space do not exceed the normal `RGB` space
- `RGB` colors should be used in [0,1], so, the clamping is the responsibility of the user

kimikage:
I'm going to add the clamping in the `HSI`-->`RGB` conversion. (cf. #407)

PR #403, opened 2020-02-17
Move promotion rules and some typealiases to ColorTypes.jl v0.10

kimikage:
~We are currently undergoing a major renovation of ColorTypes.jl. So, this PR is a preview.~

Since gray-->rgb conversions were supported and some rgb-->rgb conversions were fixed in ColorTypes.jl v0.10, this PR delegates the conversions to ColorType.jl.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -1,49 +1,51 @@
 # Conversions
 # -----------
 
-# convert(C, c) might be called as convert(RGB, c) or convert(RGB{Float32}, c)
-# This is handled in ColorTypes, which calls functions
-#     _convert(::Type{Cdest}, ::Type{Odest}, ::Type{Osrc}, c)
-#     _convert(::Type{Cdest}, ::Type{Odest}, ::Type{Osrc}, c, alpha)
-# Here are the argument types:
-# - Cdest may be any concrete Color{T,N} type. For parametric Color types
-#   it _always_ has the desired element type (e.g., Float32), so it's
-#   safe to dispatch on Cdest{T}.
-# - Odest and Osrc are Color subtypes, i.e., things like RGB
-#   or HSV. They have no element type.
-# - c is the Colorant object you wish to convert.
-# - alpha, if present, is a user-supplied alpha value (to be used in
-#   place of any default alpha or alpha present in c).
-#
-# The motivation for this arrangement is that Julia doesn't (yet) support
-# "triangular dispatch", e.g.,
-#     convert{T,C}(::Type{C{T}}, c)
-# The various arguments of _convert therefore "peel off" element types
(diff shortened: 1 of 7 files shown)


kimikage:
A concern for the releasing Colors v0.12 is https://github.com/JuliaGraphics/Colors.jl/issues/378#issuecomment-588179535.

I'm writing `normalize_hue`(cf. #379) to rewrite the conversion codes for `HSV`, `HSL` and `HSI`. Although it is simple in itself, the work involved may be time consuming.

I think the releasing should be a priority, but I don't like being called "broken".:sweat_smile:

src/conversions.jl
```diff
@@ -1,50 +1,52 @@
...
+catch all the various alpha variants (e.g. `ARGB`/`RGBA` with/without element
+type).
+The various arguments of `_convert` "peel off" element types (or guarantee them)
+so that comparisons may be made via dispatch. Therefore, this arrangement is
+still helpful.
+
+Note that ColorTypes handles the cases where `Odest == Osrc`, or they are both
+subtypes of `AbstractRGB` or `AbstractGray`. Therefore, here we only have to
+deal with conversions between different spaces.
+=#
```
kimikage: I updated the notes on the conversion scheme for the first time in four and a half years (except for renaming).
In the last four and a half years, Julia has made great strides, so I may be misunderstanding the original motivation.

cc @timholy

timholy: This looks correct, thanks for updating it! And the English is flawless.

kimikage:
TODO:
- [ ] PR #409
- [ ] PR #411
- [ ] PR #410 (optional)

It is undecided whether to eliminate the unnecessary conversion methods that are not directly related to this PR.

src/conversions.jl
```diff
@@ -713,51 +712,15 @@ end
...
-end
-function convert(::Type{ARGB32}, c::Color, alpha)
-    u = reinterpret(UInt32, convert(RGB24, c)) | (reinterpret(N0f8(alpha))%UInt32)<<24
-    reinterpret(ARGB32, u)
-end
 
 # To Gray
 # -------
+# AbstractRGB --> AbstractGray conversions are implemented in ColorTypes.jl, but
+# AbstractGray --> AbstractRGB conversions should be implemented here.
```
timholy: Don't you mean the opposite?

kimikage:
After the typo has been corrected, I will merge this PR and release Colors v0.12.

Closes #399 and Closes #414.

kimikage:
Thank you for the recent reviews, @timholy

timholy:
Thanks for doing *all* the heavy lifting, @kimikage! You can just feel the quality shining through.

Issue #398, opened 2020-01-10
Improvement of documents (phase 1)


kimikage:
The current documentation is categorized by technical field, so it is difficult for beginners to understand. For example, beginners will first read the [Colorspaces](http://juliagraphics.github.io/Colors.jl/v0.11/colorspaces/) page. However the page contains documents which most beginners do not need, such as CMFs.

Of course, adding a tutorial page is a good idea. However, I think there is still room for improvement.

The following is my plan.

## Introduction
- [x] Simplify the supported colorspaces (PR #406)
  - They are also in the "Colorspaces" page.
- [x] Add the installation instructions, i.e. `]add Colors` (PR #406)
- [x] Add notes for "reexport" (PR #406)

### Available colorspaces
- [x] Add inline figures showing the examples of 2-D cross sections (PR #406)
### :new: Constructors
- [ ] Add examples of constructors (e.g. `RGB`, `RGB{Float64}`, `RGBA`, `ARGB32`)
  - The constructors are implemented in ColorTypes.jl, so only examples are given there.
- [ ] Add notes for `Normed`
### :new: Generating random colors
- [ ] Add notes for non-RGB colors
- [ ] Add examples for sampling `RGB24` colors without replacement
  - cf. issue #434
### Color parsing
- [x] Add the explicit reference to `hex` (PR #406)
- [x] Add the link to "Named Colors" (PR #406)
### Color conversion
- [ ] Add examples with the extra arguments (e.g. `alpha`, `wp`)
## :new: Displaying colors
- [ ] Add the document about color swatches

## Color Differences (<-- Colorspaces)
- [x] Use docstrings (PR #409)
- [x] Add some graphical examples (PR #409)

### Color interpolation (<-- Generating a range of colors, Weighted color means)
- [ ] Add notes for the vector space and the reference to ColorVectorSpace.jl
- [x] Add examples for non-RGB color types. (PR #405)

#### Sequential and diverging color palettes
- [ ] Move the details to the docstrings of `sequential_palette()`/`diverging_palette`
- [ ] Add inline example figures

### Generating distinguishable colors
- [ ] Move the details to the docstring of `distinguishable_colors()`
- [ ] Add inline example swatches
- [ ] Add figures to explain the `seed`
- [x] Update the `PyPlot` example with `dropseed` (PR #415)
  - It looks like the sample numbers are different between the actual plot and the script.

## Named Colors
[...]

PR #404, opened 2020-02-19
Re-categorize docs

kimikage:
This is the first part of #398. This PR does not change the document contents.


kimikage:
I want to change the file/dir structure before releasing the next version to prevent broken links.

BTW, this PR number is 404. :ghost:

docs/make.jl
```diff
@@ -18,14 +18,15 @@ makedocs(
...
-        "Colormaps"                => "colormaps.md",
-        "Named colors"             => "namedcolors.md",
-        "References"               => "references.md",
-        "Migrating from Color.jl"  => "migratingfromcolor.md",
-        "Index"                    => "functionindex.md",
+        "Introduction"                => "index.md",
+        "Construction and Conversion" => "constructionandconversion.md",
+        "Color Differences"           => "colordifferences.md",
+        "Colormaps and Colorscales"   => "colormapsandcolorscales.md",
+        "Named Colors"                => "namedcolors.md",
```
timholy: Should "named colors" come before "colormaps and colorscales"? And maybe even before "color differences"? In a sense "named colors" is just another way of constructing colors.

docs/make.jl
```diff
@@ -18,14 +18,15 @@ makedocs(
...
+        "Advanced Functions"          => "advancedfunctions.md",
+        "References"                  => "references.md",
+        "Migrating from Color.jl"     => "migratingfromcolor.md",
```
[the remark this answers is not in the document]
kimikage: I want to freeze the file/dir structure for the time being in this PR. Let's eliminate the file.

docs/make.jl
kimikage: I think "Named Colors" should be a simple collection of color swatches and the instructions should be summarized in "Construction and Conversion". Since that will not be a simple content move, that will not be done in this PR. I think the collection of color swatches is an appendix.

cormullion: Don't like Appendices myself - too old-school pre-hyperlinked paper manuals.. :)

kimikage: @cormullion, what should I do now?

cormullion: I'd say move the section, but keeping it as it is after your improved charts. To be honest, one of the few reasons I open up this documentation is to refer to the colors - the online help is often sufficient...  But that's just my opinion, of course.

kimikage: Of course, I'm going to leave the charts, because they are part of the "tests".
I said "appendix" as to the order of the contents. I think the charts are in proper position now.

kimikage:
We have opportunities to modify the contents and their order. In other words, there is still much to be improved.:sweat_smile:

PR #406, opened 2020-02-25
Add cross-sectional charts and some examples to docs

kimikage:
This adds the cross-sectional charts previewed in https://github.com/JuliaGraphics/Colors.jl/issues/398#issuecomment-588270497.
Although there is discussion about the axis assignment (cf. https://github.com/JuliaGraphics/Colors.jl/issues/398#issuecomment-589399452), but in this PR, it is as I originally defined. Since I am currently modifying the conversion codes for `HSV`, `HSL`, and `HSI`, I just want a way to visually check the S-V, S-L and S-I planes. (**Edit:** see: #407)

In addition, this PR changes some documents.

### DIFF
--- a/docs/crosssectionalcharts.jl
+++ b/docs/crosssectionalcharts.jl
@@ -1,0 +1,259 @@
+
+module CrossSectionalCharts
+
+using Colors
+using Base64
+
+struct CrossSectionalChartSVG <: Main.SVG
+    buf::IOBuffer
+end
+
+struct Axis
+    index::Int # compornent index
+    label::String
+    range::AbstractRange
+end
+Base.first(a::Axis) = first(a.range)
+Base.last(a::Axis) = last(a.range)
+
+function crosssection(::Type{C};
+                      x::Tuple=(1, "X", 0:1),
+                      y::Tuple=(2, "Y", 0:1),
+                      z::Tuple=(3, "Z", 0:1)) where C <: Color
+    crosssection(C, Axis(x...), Axis(y...), Axis(z...))
+end
+function crosssection(::Type{C}, x::Axis, y::Axis, z::Axis) where C <: Color
+    io = IOBuffer()
+    id = String(nameof(C))
+
+    write(io,
+        """
+        <svg xmlns="http://www.w3.org/2000/svg"
+             xmlns:xlink="http://www.w3.org/1999/xlink"
+             id="svg_$id" version="1.1"
+             viewBox="0 0 40 30" width="27.09mm" height="20.32mm"
+             stroke="none" style="display:inline; margin-left:2em; margin-bottom:1em">
+        <defs>
+        <filter id="filter_g_$id" filterUnits="userSpaceOnUse" x="0" y="-0.5" width="15.5" height="16">
+          <feGaussianBlur stdDeviation="0.1" result="blur"/>
+          <feMerge>
+            <feMergeNode in="SourceGraphic"/>
+            <feMergeNode in="blur"/>
+          </feMerge>
+        </filter>
+        <style type="text/css"><![CDATA[
+          #svg_$id g, #svg_$id image {
+            transition: all 400ms ease 200ms;
+          }
+          #svg_$id path.b {
+            stroke: currentColor;
+            stroke-width: 0.5;
+            opacity: 0;
+            transition: all 200ms ease 0ms;
+          }
+          #svg_$id image {
+            opacity: 0;
+            filter:url(#filter_g_$id);
+          }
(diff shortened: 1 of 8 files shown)


docs/src/constructionandconversion.md
```diff
@@ -8,58 +8,148 @@ The colorspaces used by Colors are defined in [ColorTypes](https://github.com/Ju
...
-julia> c = colorant"red"
+julia> colorant"#f00" # 3-digit hex notation
 RGB{N0f8}(1.0,0.0,0.0)
 
-julia> parse(Colorant, "red")
+julia> colorant"rgb(255,0,0)" # rgb() notation with integers in [0, 255]
 RGB{N0f8}(1.0,0.0,0.0)
 
-julia> parse(Colorant, HSL(1, 1, 1))
-HSL{Float32}(1.0f0,1.0f0,1.0f0)
```
kimikage: I don't understand the intent of this example. I don't think this example is worth the documentation. However, this behavior is probably not what the users expect, so I included the example which pass a color to the second argument, as a warning.

docs/src/constructionandconversion.md
```diff
@@ -8,58 +8,148 @@ The colorspaces used by Colors are defined in [ColorTypes](https://github.com/Ju
...
+You can convert colors to hexadecimal strings using the [`hex`](@ref) function.
+Note that the conversion result does not have the prefix `"#"`.
+
+```jldoctest example
+julia> color = colorant"#C0FFEE"
+RGB{N0f8}(0.753,1.0,0.933)
+
+julia> hex(color)
+"C0FFEE"
+```
```
kimikage: The `hex` had been documented only with a docstring, so I mentioned it a bit. I think `hex` and `parse` should be paired.

docs/src/colordifferences.md
```diff
@@ -2,9 +2,7 @@
 
 The `colordiff` function gives an approximate value for the difference between two colors.
 
-```jldoctest example
-julia> using Colors; #hide
-
```
kimikage: This was not hidden.:sweat_smile:

docs/src/constructionandconversion.md
timholy: Agreed.

docs/src/constructionandconversion.md
```diff
@@ -8,58 +8,148 @@ The colorspaces used by Colors are defined in [ColorTypes](https://github.com/Ju
...
+or
+```jldoctest example
+julia> parse(RGB{N0f16}, "indianred")
+RGB{N0f16}(0.80392,0.36078,0.36078)
+```
+
+You can also pass a color (`Colorant` value) to the second argument of the
+[`parse`](@ref) function, but in this case, `parse` does not apply any
+conversion on the color.
```
timholy: I don't remember why we even support this. Anything bad happen if we remove it?
```julia
julia> parse(Int, 5)
ERROR: MethodError: no method matching parse(::Type{Int64}, ::Int64)
Closest candidates are:
  parse(::Type{T}, ::AbstractChar; base) where T<:Integer at parse.jl:41
  parse(::Type{T}, ::AbstractString; base) where T<:Integer at parse.jl:238
  parse(::Type{T}, ::AbstractString; kwargs...) where T<:Real at parse.jl:376
Stacktrace:
 [1] top-level scope at REPL[2]:1
```
(Perhaps Julia behaved differently at one point.)

kimikage: cf. https://github.com/JuliaGraphics/Colors.jl/issues/232

Although I think this specification is dangerous, since this PR focuses on documentation, I will keep it.

kimikage:
If you feel the charts are too small, please use the zoom function of your browser.:sweat_smile:

docs/src/constructionandconversion.md
timholy: Thanks for the link. I'd probably change my mind on that PR now.  But agreed it's a separate issue.

docs/crosssectionalcharts.jl
```diff
@@ -0,0 +1,259 @@
...
+
+
+function write_png(io::IO, cs::AbstractArray{T}) where T <: Color
+    buf = IOBuffer() # to calculate chunk CRCs
+    n = 16 # 16 x 16
+    u8(x) = write(buf, UInt8(x & 0xFF))
+    u16(x) = (u8((x & 0xFFFF)>>8); u8(x))
+    u32(x) = (u16((x & 0xFFFFFFFF)>>16); u16(x))
+    b(bstr) = write(buf, bstr)
+    function pallet(c::Color)
```
kimikage: palette...

PR #407, opened 2020-02-28
Improve HSV/HSL/HSI conversions (Fixes #378, #379)

### LINKED ISSUE #378: HSI->RGB conversion may return a value out of [0,1]

### LINKED ISSUE #379: Inconsistency in handling hues

kimikage:
~I'm working on benchmarking and tuning. So, this is a draft PR now.~ ~I want to merge #406 first to check HSx-->RGB conversions.~ (**Edit**: Done)

This fixes #378 and fixes #379.
This adds the clamping and hue normalization for sources to HSx-->RGB conversions. This also adds the clamping for destinations to HSI-->RGB conversion.

Despite the additional mechanism, at the expense of some accuracy, this speeds ​​up the conversions with some parameter combinations.

Note the following change (cf. https://github.com/JuliaGraphics/Colors.jl/issues/379#issuecomment-590213678):
```julia
julia> RGB(1,0,0)|>HSV
HSV{Float32}(0.0f0,1.0f0,1.0f0)
```

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -1,12 +1,27 @@
 # Helper data for CIE observer functions
 include("cie_data.jl")
 
+# for optimization
+div60(x) = x / 60
+_div60(x::T) where T = muladd(x, T(1/960), x * T(0x1p-6))
+if reduce(max, _div60.((90.0f0,))) == 1.5f0
+    div60(x::T) where T <: Union{Float32, Float64} = _div60(x)
+else
+    # force two-step multiplication
+    div60(x::T) where T <: Union{Float32, Float64} = x * T(0x1p-6) + x * T(1/960)
+end
+
+# mod6 supports the input `x` in [-2^28, 2^29]
+mod6(::Type{T}, x::Int32) where T = unsafe_trunc(T, x - 6 * ((widemul(x, 0x2aaaaaaa) + Int64(0x20000000)) >> 0x20))
 
 # Linear interpolation in [a, b] where x is in [0,1],
 # or coerced to be if not.
 function lerp(x, a, b)
     a + (b - a) * max(min(x, one(x)), zero(x))
 end
+
+clamp01(v::T) where {T<:Fractional} = ifelse(v < zero(T), zero(T), ifelse(v > oneunit(T), oneunit(T), v))
+clamp01(v::T) where {T<:Union{N0f8,N0f16,N0f32,N0f64}} = v
 
 """
     HexNotation{C, A, N}
@@ -148,7 +163,7 @@ Returns a normalized (wrapped-around) hue angle, or a color with the normalized
 hue, in degrees, in [0, 360]. The normalization is essentially equivalent to
 `mod(h, 360)`, but is faster at the expense of some accuracy.
 """
-@fastmath normalize_hue(h::Real) = max(fma(floor(h / 360), -360, h), zero(h))
+@fastmath normalize_hue(h::Real) = max(muladd(floor(h / 360), -360, h), zero(h))
 @fastmath normalize_hue(h::Float16) = Float16(normalize_hue(Float32(h)))
 normalize_hue(c::C) where {C <: Union{HSV, HSL, HSI}} = C(normalize_hue(c.h), c.s, comp3(c))
 normalize_hue(c::C) where {Cb <: Union{HSV, HSL, HSI}, C <: Union{AlphaColor{Cb}, ColorAlpha{Cb}}} =
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -71,7 +71,8 @@ end
 # -----------------
 
 correct_gamut(c::CV) where {CV<:AbstractRGB} = CV(clamp01(red(c)), clamp01(green(c)), clamp01(blue(c)))
-clamp01(v::T) where {T<:Fractional} = ifelse(v < zero(T), zero(T), ifelse(v > one(T), one(T), v))
+correct_gamut(c::CV) where {T<:Union{N0f8,N0f16,N0f32,N0f64},
+                            CV<:Union{AbstractRGB{T},TransparentRGB{T}}} = c
 
 function srgb_compand(v::Fractional)
     # the following is an optimization technique for `1.055v^(1/2.4) - 0.055`.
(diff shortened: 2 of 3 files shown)


kimikage:
I haven't been able to get the full performance of the CPU for now, but there are no fatal performance regressions. So, after merging PR #406, we will be ready to merge this.

src/utilities.jl
```diff
@@ -1,13 +1,28 @@
 # Helper data for CIE observer functions
 include("cie_data.jl")
 
+# for optimization
+if Sys.ARCH !== :i686
```
timholy: Does this really cover us? `fma` is really slow when implemented in software.
Perhaps we're OK though since it's a IEEE 754-2008 requirement, and 2008 was some time ago. https://en.wikipedia.org/wiki/Multiply%E2%80%93accumulate_operation#Support
But I think that's when i686 was discontinued, I think, so I'd be surprised if this comes up very often.

kimikage: In my opinion, people who use x86 processors which do not support FMA are probably not interested in speed. I have not done enough survey on ARM, but I think the high-end models support FMA.

Since Julia 1.5 is still ahead, it is possible to use `muladd` at this time.

timholy: I checked briefly, and while I don't know anything about ARM it does seem plausible that all recent models (meaning, less than 8 years old) support it. I'm good with this as the default for now, and if we get complaints we can always add new architectures.

kimikage: Note that the original purpose of this if block is not whether FMA is supported or not, but https://github.com/JuliaGraphics/Colors.jl/issues/379#issuecomment-592965144.

timholy: Gotcha. I'm happy with this!

kimikage: The alternative below does not work as expected:
```julia
_div60(x::T) where T = muladd(x, T(1/960), x * T(0x1p-6))
if _div60(90.0f0) == 1.5f0
    div60(x::T) where T <: Union{Float32, Float64} = _div60(x)
else
    # force two-step multiplication
    div60(x::T) where T <: Union{Float32, Float64} = x * T(1/960) + x * T(0x1p-6)
end
```
even though, in REPL,
```julia
julia> VERSION
v"1.5.0-DEV.376"

julia> _div60(90.0f0)
1.5000001f0

julia> div60(90.0f0)
1.5f0
```
AFAIK, only on v1.5.0-DEV `_div60` does not work as expected. However, it is not only on v1.5.0-DEV that the `if` block does not work as expected.(cf. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/131#issuecomment-547113130)

Perhaps the cause is that the stages of the constant propagation and the (LLVM IR and native) code generation are different.

kimikage: I don't think it is a bug of Julia compiler that '_div60' is drastically optimized. However, it seems that Julia drives that.
```julia
julia> @code_llvm _div60(90.0f0)

;  @ REPL[1]:1 within `_div60'
; Function Attrs: uwtable
define float @julia__div60_17301(float) #0 {
top:
; ┌ @ float.jl:404 within `*'
   %1 = fmul float %0, 1.562500e-02
; └
; ┌ @ float.jl:409 within `muladd'
   %2 = fmul contract float %0, 0x3F51111120000000
   %3 = fadd contract float %2, %1
; └
  ret float %3
}

julia> @code_llvm div60(90.0f0)

;  @ REPL[2]:5 within `div60'
; Function Attrs: uwtable
define float @julia_div60_17307(float) #0 {
top:
; ┌ @ float.jl:404 within `*'
   %1 = fmul float %0, 0x3F51111120000000
   %2 = fmul float %0, 1.562500e-02
; └
; ┌ @ float.jl:400 within `+'
   %3 = fadd float %1, %2
; └
  ret float %3
}
```
I don't know whether this is the expected behavior.

src/utilities.jl
```diff
@@ -1,13 +1,28 @@
...
 
+# for optimization
+div60(x) = x / 60
+_div60(x::T) where T = muladd(x, T(1/960), x * T(0x1p-6))
+if reduce(max, _div60.((90.0f0,))) == 1.5f0
+    div60(x::T) where T <: Union{Float32, Float64} = _div60(x)
+else
+    # force two-step multiplication
+    div60(x::T) where T <: Union{Float32, Float64} = x * T(0x1p-6) + x * T(1/960)
+end
```
kimikage: I changed the workaround for the `muladd` problem with an ad-hoc measure. This works as expected "for now".
In the nightly build, the test may fail in the future. I'll think about that at that time.  As long as the cause is clear, `===` in tests can be replaced with `≈`.

kimikage:
> https://docs.julialang.org/en/latest/devdocs/inference/#The-inlining-algorithm-(inline_worthy)-1.

I am very grateful for the sample code. I think it would be nice if a macro like `@code_inlining` was officially provided as a variation of `@code_typed`. It may be a niche and not flexible, though.

BTW, I think the cost `1` of `Base.not_int(::Bool)` is obviously too expensive. It is usually required in conditional expressions, and in most instruction sets there are instructions or special operands for negative conditions (e.g. "not equal", "not less than").  This is the answer of [the quiz](https://github.com/JuliaGraphics/Colors.jl/issues/379#issuecomment-592924172).:smile:
```julia
julia> f(x, mask) = (x & mask) != zero(x);

julia> tt = Tuple{UInt8, UInt8};

julia> mi = Base.method_instances(f, tt)[1];

julia> ci = code_typed(f, tt)[1][1]
CodeInfo(
1 ─ %1 = Base.and_int(x, mask)::UInt8
│   %2 = (%1 === 0x00)::Bool
│   %3 = Base.not_int(%2)::Bool
└──      return %3
)

julia> opt = Core.Compiler.OptimizationState(mi, Core.Compiler.Params(typemax(UInt)));

julia> cost(stmt::Expr) = Core.Compiler.statement_cost(stmt, -1, ci, opt.sptypes, opt.slottypes, opt.params);

julia> cost(stmt) = 0;

julia> for c in ci.code; @show cost(c), c;end
(cost(c), c) = (1, :(Base.and_int(_2, _3)))
(cost(c), c) = (1, :(%1 === 0x00))
(cost(c), c) = (1, :(Base.not_int(%2)))
(cost(c), c) = (0, :(return %3))
```

timholy:
I like the idea of a macro...maybe `@code_cost`?

Also, I'm glad for your input about the costs themselves. The precise numbers I picked, while within stated ranges, were pretty arbitrary. Perhaps a better strategy would be to have an architecture-dependent tuning process, but I didn't go to that kind of effort. If there's a clear case for fixing some of them we should just do that.

src/utilities.jl
timholy: Sounds good to me!

kimikage:
This is off-topic, but I implemented the `@code_costs`. There is no sense of unity in the naming. (`@code_llvm` is using a "noun", though.)
https://github.com/kimikage/CodeCosts.jl

PR #408, opened 2020-02-28
Re-enable clamping in `hex`

kimikage:
See: https://github.com/JuliaGraphics/Colors.jl/issues/378#issuecomment-588179535

I had forgotten the alpha outside of [0,1].  Although some slowing down is inevitable, it is still faster than pre-PR #387.


timholy:
I can go either way on this (either do it automatically or force the user to call clamp).

kimikage:
I think the automatic clamping is a hindrance to gamut conversions. However, considering other use cases, I think it is safe to do clamping automatically.

PR #409, opened 2020-03-03
Add charts for `colordiff` to docs and Use docstrings for metrics

kimikage:
This update the docs for color differences (cf. https://github.com/JuliaGraphics/Colors.jl/issues/398#issuecomment-591360224).

I changed the example for `colordiff`, i.e. removed the strange `parse`.

I also add a doctest for the difference between an `HSV` color and an `HSL` color. It may help check the PR #407. ~The PR #407 is ready for merging, but there may be room for improvement on the `fma`/`muladd` (cf. https://github.com/JuliaGraphics/Colors.jl/pull/407#discussion_r386098781). So, it is another reasonable option to merge this PR first.~ ~(**Edit:** Either is fine now.)~
PR #407 was merged.

PR #411, opened 2020-03-04
Add abstract type EuclideanDifferenceMetric

kimikage:
This simplifies the Euclidean difference metrics.

I am planning to clean up "differences.jl", but it is likely that there will be many fixes but no breaking changes. Therefore, I extracted only the part related to the type hierarchy. This PR is the successor to PR #409. (**Edit:** PR #409 was merged.)

### DIFF
--- a/src/differences.jl
+++ b/src/differences.jl
@@ -2,6 +2,8 @@
 # Define an abstract type to represent color difference metrics
 abstract type DifferenceMetric end
 
+abstract type EuclideanDifferenceMetric{T<:Color3} <: DifferenceMetric end
+
 # TODO?: make the DifferenMetrics parametric, to preserve type-stability
 
 struct DE_2000 <: DifferenceMetric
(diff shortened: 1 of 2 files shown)


src/differences.jl
```diff
@@ -2,6 +2,8 @@
 # Define an abstract type to represent color difference metrics
 abstract type DifferenceMetric end
 
+abstract type EuclidianDifferenceMetric{T<:Color3} <: DifferenceMetric end
```
kimikage: Of course, there is no need to constrain the parameter to `Color3`, but currently only `Color3` is needed.
I think it's okay to relax the constraint in the future.

johnnychen94: Any reference for the name `Euclidian` (or `Euclidean`)?

kimikage: I assumed that its spelling was "Euclidian", but "Euclidean" has been used in the docs.
http://juliagraphics.github.io/Colors.jl/stable/colorspaces/#Color-Difference-1

Perhaps it's not a term in chromatics, but I think the name is reasonable for the internal (unexported) abstract type.

Issue #372, opened 2019-12-08
Problem with conversion of `YIQ` (or problem with non-sRGB gamut)


kimikage:
I have a proposal to add abstract gamut types in `ColorTypes.jl`.(https://github.com/JuliaGraphics/ColorTypes.jl/pull/140)
I think it will help reduce the misunderstandings about the color gamut and the misfortunes from them.

However, it also clarifies the issues that have been obscured so far. One of them is a problem with conversion of `YIQ`.

> The color primaries of NTSC and sRGB are different, and the NTSC color gamut is actually wider than sRGB's one.

_Originally posted by @kimikage in https://github.com/JuliaGraphics/ColorTypes.jl/pull/140#discussion_r355152750

In fact, the current `RGB` is *implicitly* considered as "sRGB". On the other hand, the current conversion methods from/to `YIQ` consider that the `RGB` is defined under the "NTSC" gamut.

https://github.com/JuliaGraphics/Colors.jl/blob/1bc194e5e2122b5c9e69f9d55e12afc1ad0e9759/src/conversions.jl#L712-L723

The trouble is that the gamuts are often ignored when we use `YIQ` or `YCbCr`. In fact, some image processing tools do such conversions which are inaccurate from a color management perspective. However, in the testing, the inaccuracy or inconsistency is a big problem.

kimikage:
I don't think `YIQ` is used so often. Therefore, it may be a good idea to add the gamut conversion to the conversion methods, and add low-level conversion methods like `rgb_to_yiq_ntsc` and `yiq_to_rgb_ntsc`.

Again, this is because I think that `YIQ` is not used much. The color gamut issues potentially have a similar problem to the white points.
https://github.com/JuliaGraphics/Colors.jl/issues/278#issuecomment-537755564

kimikage:
cf. #8

kimikage:
I will change the gamut for `YIQ` to "SMPTE-C" instead of "sRGB" or (original) "NTSC".
We can use `yiq_to_rgb_smptec`/`rgb_to_yiq_smptec` as the compatible methods without gamut conversion.

However, the gamma and the narrow sense of the color gamut are independent concepts. There is already a lot of confusion. :confused:

kimikage:
I decided not to change this unless anyone else wanted the fix in Colors.jl.
cf. https://github.com/JuliaGraphics/Colors.jl/pull/412#issuecomment-602595025

kimikage:
[...]

PR #412, opened 2020-03-05
[RFC/WIP] Add partial support for color gamuts

kimikage:
This is a preview for the gamuts support. (cf. #372, https://github.com/JuliaGraphics/ColorTypes.jl/pull/140)

Functions for so-called color management will not be implemented for the time being.
I'm going to use them only internally (i.e. without exporting them) as experimental features.

From the experience of the experimental implementation of the chromatic adaptation (#340), I think designing APIs for conversion is quite difficult.

### DIFF
--- a/src/Colors.jl
+++ b/src/Colors.jl
@@ -25,6 +25,7 @@ export weighted_color_mean,
 include("utilities.jl")
 
 # Include other module components
+include("gamuts.jl")
 include("conversions.jl")
 include("algorithms.jl")
 include("parse.jl")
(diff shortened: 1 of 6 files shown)


timholy:
I'm not going to pretend I can give you a good review here without investing far more than I currently have time for. Would be best if one of the colorimetry folks reviewed it.

kimikage:
I have been skeptical about the benefits of making such major changes in Colors.jl since late last year, i.e. when we started on the road to v1.0.
Now, I think it's a good thing to modularize functionalities. In other words, I think it is better to create new packages to supplement Colors.jl than to change Colors.jl itself.
However, I'm not sure it is a really good way.

PR #373, opened 2019-12-09
update logo to modernized Julia

cormullion:
Screenshot 2019-12-09 at 09 54 04

PR #413, opened 2020-03-08
Add example images of simulation of color deficiency

kimikage:
cf. https://github.com/JuliaGraphics/Colors.jl/issues/398#issuecomment-596120539


kimikage:
The sample image included in this PR is the tricky  version with reduced colors and sizes. The original version is as follows (CC0).
beads_1920x1080

johnnychen94:
I like the colors 😍

kimikage:
> may I add this to TestImages?

Sure. It is released under the Creative Commons CC0 license.

kimikage:
I forgot to erase unnecessary code, so I removed it.:sweat_smile:

Since this changes the docs only, I will merge this within a few days if there is no objection.

timholy:
It seems likely to be useful for many purposes, not just colors but segmentation algorithms, algorithms inferring depth from blur, etc. And it's just plain beautiful. Thanks, @kimikage!

Out of curiosity, why is it an svg? Did you actually generate this image rather that it being a photograph?

kimikage:
> Out of curiosity, why is it an svg? Did you actually generate this image rather that it being a photograph?

This is a 3d computer-generated image, and post-processed to make some glass beads to be in indistinguishable color pairs.

I hated depending on ImageMagick just for these examples, so I decided to use the technique used in the cross-sectional charts (cf. PR #406), i.e. the use of paletted PNGs. However, 256 colors are too few to represent this image.
And here's the thing. :smile:

beads_y
beads_g
beads_r

We shouldn't really use the dithering because of the different color spaces from normal vision. Therefore, I adjusted the hues so that they did not deviate significantly.

kimikage:
I made a mistake on the HTML syntax. :sob:
I don't know whether the cause is it or the embedded PNG, but the github pages hasn't been updated.

docs/sampleimages.jl
```diff
@@ -0,0 +1,93 @@
...
+function BeadsImageSVG(caption::String; filter=nothing, width="64mm", height="36mm")
+    io = IOBuffer()
+    id = string(hash(caption), base=16)
+
+    open(joinpath("assets", "figures", "beads.svg"), "r") do file
+        for line in eachline(file, keep=true)
+            occursin("<?xml", line) && continue
+            if occursin("<svg", line)
+                line = replace(line, ">"=>
+                    """width="$width" height="$height" style="display:inline; margin-left:1em; margin-bottom:1em">""")
```
kimikage:
```diff
-"""width="$width"
+""" width="$width"
```

kimikage:
Updated! :smile:
http://juliagraphics.github.io/Colors.jl/latest/advancedfunctions/
(triggerd by PR #373)

timholy:
The new docs just look fantastic.

johnnychen94:
FYI, Documenter 0.24 provides a `push_preview` keyword for non-fork PRs

For example, https://github.com/JuliaImages/juliaimages.github.io/pull/115

Issue #298, opened 2017-10-30
A way to make distinguishable_colors output darker, for use on white backgrounds?


ianshmean:
I like using `distinguishable_colors` with PyPlot, but find that with thin line thicknesses, some of the the colors wash out on a white background (see yellow below). Is there a simple way to weight the process towards darker colors?

```
using PyPlot, Colors
vars = 1:10
cols = distinguishable_colors(length(vars))
plot_cols = map(col->(red(col),green(col),blue(col)),cols)
for i = 1:length(vars)
    plot(1:10,rand(10),"-o",c = plot_cols[i],linewidth=0.5,markersize=1)
end
```
![image](https://user-images.githubusercontent.com/1694067/32179608-3ae2ad1e-bd66-11e7-8ca4-6336f01c7475.png)

timholy:
Supply white as a seed, and then drop it from the returned list of colors:
```julia
julia> distinguishable_colors(3)
3-element Array{ColorTypes.RGB{FixedPointNumbers.Normed{UInt8,8}},1}:
 RGB{N0f8}(0.0,0.0,0.0)  
 RGB{N0f8}(1.0,1.0,0.384)
 RGB{N0f8}(1.0,0.624,1.0)

julia> distinguishable_colors(4, [RGB(1,1,1)])
4-element Array{ColorTypes.RGB{FixedPointNumbers.Normed{UInt8,8}},1}:
 RGB{N0f8}(1.0,1.0,1.0)      
 RGB{N0f8}(0.0,0.0,0.0)      
 RGB{N0f8}(0.757,0.161,0.886)
 RGB{N0f8}(0.82,0.29,0.0)    
```

ianshmean:
Ah, of course, thanks. Perhaps a hint in docs might be helpful, giving the explicit example of use with pyplot?
```
cols = distinguishable_colors(length(variants)+1,[RGB(1,1,1)])[2:end]
pcols = map(col->(red(col),green(col),blue(col)),cols) 
```

timholy:
Could you add that? Just open the file on GitHub and click the pencil icon.

ianshmean:
Done. #299

tpapp:
I am using this so much that I am wondering if a `dropseed` keyword argument could be added.

PR #415, opened 2020-04-08
Update PyPlot example

kimikage:
This changes the image format from PNG to SVG, and makes its colors dynamically set.


kimikage:
> It looks like the sample numbers are different between the actual plot and the script.

and it is 0.9MB! :sweat_smile:
```
docs/src/assets/figures/pyplot-seed-dcols.png PNG 1748x1026 1748x1026+0+0 8-bit sRGB 874257B
```

Of course, it is another option to generate the plot dynamically using `PyPlot`. However, the technical points of that example are to use the `seed` option and to convert the colors to arrays (cf. https://github.com/JuliaGraphics/Colors.jl/issues/298#issuecomment-340563024), so I don't think the plotting result other than its colors is important.

Issue #361, opened 2019-11-22
Road to 1.0


timholy:
I'd like to move the whole JuliaImages stack to version numbers >= 1.0 (see JuliaImages/Images.jl#825). Let's use this issue to collect plans for breaking changes in the near future.

Issues/PRs on my agenda:
- #360 
- #353
- possibly #349, but I worry that will take a while and hence might be 2.0 material
- #273 (ugh)

kimikage:
- A part of #354?

Actually, I have forgotten how far I have worked :sweat_smile:, but somethings should have already been achieved.
https://github.com/JuliaGraphics/Colors.jl/compare/master...kimikage:cnvt

However, there are few breaking changes other than the bug fixes. In regards to the white point `wp`,it follows the current interface, but it is necessary to discuss how white points should be handled.

timholy:
You have done wonders for these packages.

"Cleanups" can be done after 1.0 if they don't break the public API. If we grossly change the result of a white-point aware conversion, though, that is certainly breaking.

kimikage:
- #371 following #353

kimikage:
CC @cormullion
- Tidy the logo up (cf. https://github.com/JuliaLang/julia-logo-graphics)

It may be useful for the testing to add "julia_blue", "julia_green", "julia_purple" and "julia_red" to the named colors. I also feel that "nationalism" is too much, though. :smile:

cormullion:
No problem. What do you mean by "I also feel that "nationalism" is too much, though."?

kimikage:
Adding "julia_blue" etc. to named colors means that it is not necessarily a *public* interest.
"Colors of the Julia users, by the Julia users, for the Julia users!"

cormullion:
Ah, I see. :) Anyway, probably not many people use them (apart from me, I expect)...

kimikage:
Reminder: https://github.com/JuliaGraphics/Colors.jl/issues/353#issuecomment-570967804

kimikage:
- [ ] #496

jonathanBieler:
If there breaking changes in master currently ? I wanted to use some of the XYZ to RGB conversion but last release was before it was added (27 Apr 2021). Would it be possible to make minor release before 1.0 ?

timholy:
Yeah, we need to do something about this. I am pretty swamped with TTFX work until at least mid-October, unfortunately. If you'd be willing to figure out what to do perhaps we could review your work and then make a release.

[...]

PR #418, opened 2020-04-22
Add a constant for the colors in the Julia logo

tpapp:
The colors from the Julia logo are frequently used for various visualizations and logo designs, private or public, so I thought they would be useful in this package.

At the same time, I did not put them in the named colors namespace since they are unique to Julia, not a shared color name space like X11 or SVG (cf [this comment](https://github.com/JuliaGraphics/Colors.jl/issues/361#issuecomment-563033679) by @kimikage). For the same reason, they are not exported, just mentioned in the docs.

Happy to make any changes.

(RGB color codes just extracted from the svg in the JuliaLang/julia repository with Inkscape).

### DIFF
--- a/docs/src/namedcolors.md
+++ b/docs/src/namedcolors.md
@@ -65,6 +65,6 @@ NamedColorCharts.ColorChartSVG("grays") # hide
 
 ## Julia logo colors
 
-```docs
+```@docs
 Colors.JULIA_LOGO_COLORS
 ```
--- a/docs/src/namedcolors.md
+++ b/docs/src/namedcolors.md
@@ -62,3 +62,9 @@ NamedColorCharts.ColorChartSVG("grays") # hide
     named colors. There are some unnatural definitions due to the different
     origins. For example, "LightGray" is lighter than "Gray", but "DarkGray" is
     also lighter than "Gray".
+
+## Julia logo colors
+
+```docs
+Colors.JULIA_LOGO_COLORS
+```
--- a/docs/src/namedcolors.md
+++ b/docs/src/namedcolors.md
@@ -64,7 +64,14 @@ NamedColorCharts.ColorChartSVG("grays") # hide
     also lighter than "Gray".
 
 ## Julia logo colors
-
+[`Colors.JULIA_LOGO_COLORS`](@ref) is a `NamedTuple` containing the julia logo
+colors defined in the [julia-logo-graphics](https://github.com/JuliaLang/julia-logo-graphics#color-definitions).
+```@example julialogo
+using Colors # hide
+logocolors = Colors.JULIA_LOGO_COLORS
+showable(::MIME"text/plain", ::AbstractVector{C}) where {C<:Colorant} = false # hide
+[logocolors.blue, logocolors.red, logocolors.green, logocolors.purple]
+```
 ```@docs
 Colors.JULIA_LOGO_COLORS
 ```
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -234,7 +234,7 @@ The keys are approximate descriptions of the hue and do not include black.
 
 Not exported, use as `JULIA_LOGO_COLORS.red` etc.
 """
-const JULIA_LOGO_COLORS = (red = colorant"#ca3c32",
-                           green = colorant"#399746",
-                           blue = colorant"#4d64ae",
-                           purple = colorant"#9259a3")
+const JULIA_LOGO_COLORS = (red = colorant"#cb3c33",
+                           green = colorant"#389826",
+                           blue = colorant"#4063d8",
+                           purple = colorant"#9558b2")
--- a/src/parse.jl
+++ b/src/parse.jl
@@ -226,3 +226,15 @@ end
     Base.depwarn("color(\"$str\") is deprecated, use colorant\"$str\" or parse(Colorant, \"$str\")", :color)
     parse(Colorant, str)
 end
+
+"""
+Colors used in the Julia logo as a `NamedTuple`.
+
+The keys are approximate descriptions of the hue and do not include black.
+
+Not exported, use as `JULIA_LOGO_COLORS.red` etc.
+"""
+const JULIA_LOGO_COLORS = (red = colorant"#ca3c32",
+                           green = colorant"#399746",
+                           blue = colorant"#4d64ae",
+                           purple = colorant"#9259a3")


kimikage:
Thank you for submitting this PR and your consideration for the cleanness of namespace.

> (RGB color codes just extracted from the svg in the JuliaLang/julia repository with Inkscape).

The official definitions of the colors are found in https://github.com/JuliaLang/julia-logo-graphics. The difference is weird. :thinking:
In any case, we should use the official definitions.

My concern is whether `JULIA_LOGO_COLORS` can be easily recalled. It's not clear which of the Colors.jl documentation or the site above is easier to access. :sweat_smile:

If we don't add the colors to the named colors, what about adding the `juliacolor_str` macro (e.g. `juliacolor"red"`)? The string literal macro seems unlikely to cause name conflicts, so IMO, exporting the macro may be OK.

cc: @cormullion

cormullion:
It's a great idea! I think `JULIA_LOGO_COLORS.red` etc seems a pretty good way to get the colors.

The color specifications currently originate from Luxor.jl:

```
julia> Luxor.julia_red, Luxor.julia_green, Luxor.julia_blue, Luxor.julia_purple
((0.796, 0.235, 0.2), (0.22, 0.596, 0.149), (0.251, 0.388, 0.847), (0.584, 0.345, 0.698))

julia> JULIA_LOGO_COLORS
(red = RGB{N0f8}(0.792,0.235,0.196), green = RGB{N0f8}(0.224,0.592,0.275), blue = RGB{N0f8}(0.302,0.392,0.682), purple = RGB{N0f8}(0.573,0.349,0.639))
```

so they're pretty close (I probably couldn't spot the difference, but I suspect you can, @kimikage :)).

(The "julia-logo-graphics" repo is _kind of_ official, now that it's been transferred from my github to JuliaLang. But there was never really any official specification of the logo or colors before then - the logo image just travelled around in various formats, and I eventually decided to pin it down and draw it procedurally.)

tpapp:
@kimikage: I checked and the colors match the ones in [julia.svg](https://github.com/JuliaLang/julia/blob/ac825137958bf0bd4cfd61b3722622c8880146c0/contrib/julia.svg) in the latest master. I checked the repo you linked and it is my understanding that it just draws colors from Luxor.jl --- while I like that package very much, I am not sure in what sense it is more official than the Julia repository. Naturally, if the colors are updated in the latter, that should be reflected here.

The purpose of this PR is to make these colors easier to access when needed for Julia programs. I am not sure in what sense you consider the repo above an alternative; the colors are not accessible in code directly, they would need to be typed in from an *image* in that repository, which is tedious and error-prone.

In order to avoid getting into the "which colors are official" debate, perhaps I should amend the docstring to say that it should just reflect whatever is in the Julia repository, and kept updated accordingly, without considering these changes breaking.

tpapp:
Regarding the macro: I just prefer a constant. We are talking about a very small set of values after all, for which something like a `NamedTuple` is ideal, providing completion etc out of the box.

cormullion:
The julia.svg in the main JuliaLang repository was created by Luxor... 🤷. I expect the differences in values are due to different ways of converting the original 0 -> 1.0 values to those hexadecimal codes.

kimikage:
As I want to use these colors ​​for testing, I don't care which one is "official", but the existence of multiple possible answers is a concern.:confused: 

We should consider the possibility of the unification before merging this PR. And this, first of all, the "copyright owner" should have the first say.

I also think we should clarify the reason for the variation. To be precise, the cause itself is not so important, but we should make sure that the same problem does not recur. Some differences can be explained as rounding errors, but when it comes to "blue", they are clearly different colors.


> Regarding the macro: I just prefer a constant. 

Adding macros does not mean removing constants. We need a dictionary-like object to convert from strings to colors. However, the macro is just a possible idea.

kimikage:
@cormullion, what happened in round2? It seems that the result of round2 is not the raw output of Luxor.
Round1 (https://github.com/JuliaLang/www.julialang.org/pull/372)
https://github.com/JuliaLang/www.julialang.org/blob/9c2ffde0233da26f7d5f1e07408823825860d46b/_includes/logo.svg

Round2 (https://github.com/JuliaLang/julia/pull/32038)
https://github.com/JuliaLang/julia/blob/b5a4601bbe8a6f1b8ce8f4e307543850c08adb8e/contrib/julia.svg

IMO, https://github.com/JuliaLang/julia/blob/master/contrib/julia.svg should be changed.

cormullion:
"what happened in round2" - I can't remember now, but I must have used the wrong file at some point.

tpapp:
Frankly, I expected this PR to be rather trivial (conditional on the functionality being considered to have a place in this package), and I am somewhat surprised about the whole "which colors are official" discussion, with the claim that some other colors are, because they are used in a package that was ostensibly used to create the logo, which nevertheless ended up having different colors.

The point of the whole PR is to just make using some colors that some people use anyway easier. I don't wish to invest too much in this, and I don't see the point of waiting for a resolution between the semi-official logo repo and the svg file. If/when that happens, colors here can change too.

In any case, feel free to ignore/close/modify this PR.

kimikage:
At least the discussion of "which color is official" ended. The current discussion is about "official" colors, not unofficial or semi-official. So, it's not appropriate to have that discussion in this PR or in Colors.jl.

kimikage:
As the log of Travis CI shows, the document cannot be built due to the specification change of nightly build now. Although problem can be solved by ignoring the failure, I am currently exploring best practices. In the meantime, I leave this PR open.

BTW, I want to add the swatches of the colors to the document.
By suppressing the display in "text/plain", showing swatches in SVG is enabled.
```julia
showable(::MIME"text/plain", ::AbstractVector{C}) where {C<:Colorant} = false # hide
```

tpapp:
@kimikage, @cormullion: thanks for clearing this up. I have updated the colors accordingly.

docs/src/namedcolors.md
```diff
@@ -62,3 +62,9 @@ NamedColorCharts.ColorChartSVG("grays") # hide
     named colors. There are some unnatural definitions due to the different
     origins. For example, "LightGray" is lighter than "Gray", but "DarkGray" is
     also lighter than "Gray".
+
+## Julia logo colors
+
+```docs
```
kimikage, line 68:
```suggestion
```@docs
```

docs/src/namedcolors.md
```diff
@@ -62,3 +62,9 @@ NamedColorCharts.ColorChartSVG("grays") # hide
     named colors. There are some unnatural definitions due to the different
     origins. For example, "LightGray" is lighter than "Gray", but "DarkGray" is
     also lighter than "Gray".
+
+## Julia logo colors
+
```
kimikage, line 67:
````suggestion
[`Colors.JULIA_LOGO_COLORS`](@ref) is a `NamedTuple` containing the julia logo
colors defined in the [julia-logo-graphics](https://github.com/JuliaLang/julia-logo-graphics#color-definitions).
```@example julialogo
using Colors # hide
logocolors = Colors.JULIA_LOGO_COLORS
showable(::MIME"text/plain", ::AbstractVector{C}) where {C<:Colorant} = false # hide
[logocolors.blue, logocolors.red, logocolors.green, logocolors.purple]
```
````

kimikage:
Thanks!

Issue #419, opened 2020-04-23
[RFC] Stopping use of depwarn intended to inform to "end users"


kimikage:
In Julia v1.5 series (or nightly build),
> * Deprecation warnings are no longer shown by default. i.e. if the `--depwarn=...` flag is not passed it defaults to `--depwarn=no`.

cf. https://github.com/JuliaLang/julia/pull/35362

Therefore, depwarns will be for package developers, not for end users. (We use depwarn instead of normal warn just because it's annoying.:sweat:)

There are 4 depwarns used in this package.

## Base.hex
https://github.com/JuliaGraphics/Colors.jl/blob/a447509004d10099b8ee4445dfdb8fdc097dec61/src/utilities.jl#L103-L106
Perhaps the end users won't explicitly specify `Base`, so we can leave this. (We can also remove this.)

## hex
https://github.com/JuliaGraphics/Colors.jl/blob/a447509004d10099b8ee4445dfdb8fdc097dec61/src/utilities.jl#L109-L116
This is a problem as the new behavior will not involve any API changes. Also, `hex` is a function used directly by the end users. However, I don't think we need to show this depwarn to the end user when Julia v1.5 is released. So, we can leave this.

## ColorTypes.color
https://github.com/JuliaGraphics/Colors.jl/blob/f1047dafb88f72ef242cfc71b6358b9460562835/src/parse.jl#L225-L228
This is a proper use case of depwarn, but I don't think it's necessary anymore. Moreover, `color` is defined in ColorTypes and exported (and re-exported).

## parse
https://github.com/JuliaGraphics/Colors.jl/blob/f1047dafb88f72ef242cfc71b6358b9460562835/src/parse.jl#L143-L147

This has been a problem since the beginning. (cf. https://github.com/JuliaGraphics/Colors.jl/pull/390)
This is mainly for end users, so it doesn't make sense if it disappears. At least this is what should be displayed as a normal warn.

kimikage:
This is off-topic but I don't understand CI failures on nightly build.
```julia
Parse: Log Test Failed at /home/travis/build/JuliaGraphics/Colors.jl/test/parse.jl:17
  Expression: parse(Colorant, "sea GREEN")
  Log Pattern: (:warn, r"Use \"SeaGreen\" or \"seagreen\"")
  Captured Logs: 
```
`Pkg.test` uses `--depwarn=yes` now (cf. https://github.com/JuliaLang/Pkg.jl/pull/1763).

[...]

PR #420, opened 2020-04-24
Use @test_deprecated for depwarn tests

kimikage:
In Julia v1.5 series, deprecation warnings are no longer shown by default. Although `Pkg.test` will soon enable the display of deprecation warnings, this changes the method for testing deprecation warnings to make the tests robust.

Apart from that, I think we need to rethink the usage of deprecation warnings (cf. #419).


kimikage:
I think there is no need to rush #419. This PR also

kimikage:
You can reopen this freely.

PR #421, opened 2020-04-28
Fix error types and conditions in `weighted_color_mean`

### LINKED ISSUE #416: `DomainError` needs a val argument.
https://github.com/JuliaGraphics/Colors.jl/blob/a447509004d10099b8ee4445dfdb8fdc097dec61/src/utilities.jl#L208

https://github.com/JuliaGraphics/Colors.jl/blob/a447509004d10099b8ee4445dfdb8fdc097dec61/src/utilities.jl#L212

kimikage:
This fixes #416, but leaves the weight type conversion (https://github.com/JuliaGraphics/Colors.jl/issues/416#issuecomment-619586809).

`Gray{Bool}`s do not throw a `DomainError` due to their type now, although the `InexactError` is likely to occur. Because of the weight type conversion, the following throws an `InexactError`.
```julia
julia> weighted_color_mean(0.5, Gray{Bool}(1), Gray{Bool}(1))
ERROR: InexactError: Bool(0.5)
```

The range checking of `w1` is done in the type of `w1` in order to more intuitive error messages.
#### Before (master):
```julia
julia> weighted_color_mean(-0.001, Gray24(0), Gray24(1))
ERROR: ArgumentError: FixedPointNumbers.Normed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0; cannot represent -0.001

julia> weighted_color_mean(1.00000005, Gray(0.0f0), Gray(1.0f0)) # Float32(1.00000005) == 1.0f0
Gray{Float32}(0.0f0)
```
#### After (this PR):
```julia
julia> weighted_color_mean(-0.001, Gray24(0), Gray24(1))
ERROR: DomainError with -0.001:
`w1` must be in [0, 1]

julia> weighted_color_mean(1.00000005, Gray(0.0f0), Gray(1.0f0))
ERROR: DomainError with 1.00000005:
`w1` must be in [0, 1]
```

I'm not sure wheter this is a breaking change. If this is a breaking change, I think it is better to remove the weight type conversion in this PR.

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -203,14 +203,16 @@ function weighted_color_mean(w1::Real, c1::C, c2::C) where {Cb <: Union{HSV, HSL
                                                             C <: Union{Cb, AlphaColor{Cb}, ColorAlpha{Cb}}}
     normalize_hue(_weighted_color_mean(w1, c1, c2))
 end
-function weighted_color_mean(w1::Real, c1::Gray{Bool}, c2::Gray{Bool})
-    # weighting of two Gray{Bool} would return different color type and therefore omitted
-    throw(DomainError())
-end
 function _weighted_color_mean(w1::Real, c1::Colorant, c2::Colorant)
-    weight1 = convert(promote_type(eltype(c1), eltype(c2)),w1)
-    weight2 = weight1 >= 0 && weight1 <= 1 ? oftype(weight1,1-weight1) : throw(DomainError())
-    mapc((x,y)->weight1*x+weight2*y, c1, c2)
+    @fastmath min(w1, oneunit(w1) - w1) >= zero(w1) || throw(DomainError(w1, "`w1` must be in [0, 1]"))
+    T = promote_type(eltype(c1), eltype(c2))
+    weight1 = convert(T, w1) # TODO: Consider the need for this
+    weight2 = oneunit(weight1) - weight1
+    mapc((x, y)->convert(T, muladd(weight1, x, weight2 * y)), c1, c2)
+end
+function _weighted_color_mean(w1::Integer, c1::C, c2::C) where C <: Colorant
+    (w1 & 0b1) === w1 || throw(DomainError(w1, "`w1` must be in [0, 1]"))
+    w1 == zero(w1) ? c2 : c1
 end
 
 """
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -167,43 +167,59 @@ using Colors, FixedPointNumbers, Test, InteractiveUtils
     colortypes = vcat(parametric3,parametric4A,parametricA4)
     colorElementTypes = [Float16,Float32,Float64,BigFloat,N0f8,N6f10,N4f12,N2f14,N0f16]
 
-    for T in colorElementTypes
-        c1 = Gray{T}(1)
-        c2 = Gray{T}(0)
-        @test weighted_color_mean(0.5,c1,c2) == Gray{T}(0.5)
-    end
-
-    for C in parametric2
+    @testset "weighted_color_mean" begin
         for T in colorElementTypes
-            C == AGray32 && (T=N0f8)
-        c1 = C(T(0),T(1))
-            c2 = C(T(1),T(0))
-        @test weighted_color_mean(0.5,c1,c2) == C(T(1)-T(0.5),T(0.5))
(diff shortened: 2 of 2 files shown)


src/utilities.jl
```diff
@@ -203,14 +203,16 @@ function weighted_color_mean(w1::Real, c1::C, c2::C) where {Cb <: Union{HSV, HSL
...
-    weight1 = convert(promote_type(eltype(c1), eltype(c2)),w1)
-    weight2 = weight1 >= 0 && weight1 <= 1 ? oftype(weight1,1-weight1) : throw(DomainError())
-    mapc((x,y)->weight1*x+weight2*y, c1, c2)
+    @fastmath min(w1, oneunit(w1) - w1) >= zero(w1) || throw(DomainError(w1, "`w1` must be in [0, 1]"))
+    T = promote_type(eltype(c1), eltype(c2))
+    weight1 = convert(T, w1) # TODO: Consider the need for this
+    weight2 = oneunit(weight1) - weight1
+    mapc((x, y)->convert(T, muladd(weight1, x, weight2 * y)), c1, c2)
+end
+function _weighted_color_mean(w1::Integer, c1::C, c2::C) where C <: Colorant
```
johnnychen94, line 213: Any reason not allowing different types of `c1` and `c2`, since it only requires `w1` to be `0` or `1`.

```suggestion
function _weighted_color_mean(w1::Integer, c1::Colorant, c2::Colorant)
```

src/utilities.jl
```diff
@@ -203,14 +203,16 @@ function weighted_color_mean(w1::Real, c1::C, c2::C) where {Cb <: Union{HSV, HSL
...
-    # weighting of two Gray{Bool} would return different color type and therefore omitted
-    throw(DomainError())
-end
 function _weighted_color_mean(w1::Real, c1::Colorant, c2::Colorant)
-    weight1 = convert(promote_type(eltype(c1), eltype(c2)),w1)
-    weight2 = weight1 >= 0 && weight1 <= 1 ? oftype(weight1,1-weight1) : throw(DomainError())
-    mapc((x,y)->weight1*x+weight2*y, c1, c2)
+    @fastmath min(w1, oneunit(w1) - w1) >= zero(w1) || throw(DomainError(w1, "`w1` must be in [0, 1]"))
+    T = promote_type(eltype(c1), eltype(c2))
+    weight1 = convert(T, w1) # TODO: Consider the need for this
```
johnnychen94, line 209: I guess we should make the following work without throwing error?

```julia
weighted_color_mean(0.5, Gray{Bool}(0), Gray{Bool}(0))
```

src/utilities.jl
kimikage: Any reason allowing different types of `c1` and `c2`? They cannot be optimized, i.e. can throw errors.

src/utilities.jl
kimikage: I noticed this issue when tackling the problem with `Gray{Bool}`, but I think `RGB{N0f8}` is more important for people involved in image processing.

johnnychen94: I mean this could be a good reason to remove `convert(T, w1)` 😃

src/utilities.jl
johnnychen94: I don't get it, why would it throw errors? Is that because `c1` and `c2` are not promoted to a common type?

kimikage: > Is that because c1 and c2 are not promoted to a common type?

Yes. Of course, no error should be thrown for colors in the same color space, but this optimization is not available.

```julia
julia> weighted_color_mean(1, Gray{Float32}(0), Gray{Float64}(1))
Gray{Float64}(0.0)
```

src/utilities.jl
kimikage: I agree with you. However, I don't understand the purpose of the conversion, so I can't come to a conclusion. Whatever the reason, if we remove it, it should be in v0.13.

kimikage: The removal requires some modifications of tests, so I think it's better to make a separate PR. If we make the two steps anyway, it's a bit safer to do this in v0.12 and the removal in v0.13.

kimikage:
Thank you for your review despite your busy schedule, @timholy.
Also, thank you for your comment, @johnnychen94.

I opened a new issue for removing the conversion. I will submit a PR later.

Issue #425, opened 2020-05-23
Consideration of workarounds for `cbrt` and `exp` precompilation problems


kimikage:
**Edit:** For the `exp` case, a workaround has been introduced in PR #483.

As https://github.com/JuliaLang/julia/issues/35972 revealed, `cbrt` and `exp` have problems with precompilation.

In the Colors.jl, `cbrt` and `exp` are used in `XYZ`-->{`Lab`, `Luv`} and `XYZ` <--> `RGB` conversions.

https://github.com/JuliaGraphics/Colors.jl/blob/bea98d9d8c64142795c36145b5a3171b0d065c3c/src/conversions.jl#L430-L436

https://github.com/JuliaGraphics/Colors.jl/blob/bea98d9d8c64142795c36145b5a3171b0d065c3c/src/conversions.jl#L525-L536

https://github.com/JuliaGraphics/Colors.jl/blob/bea98d9d8c64142795c36145b5a3171b0d065c3c/src/utilities.jl#L21-L27

Note that the conversion methods from the `DIN99` and its variants also use `exp` but they don't cause this problem since they are not precompiled.

`RGB{Float32}`->`XYZ` and `RGB{N0f8}`->`XYZ` are also OK because of the specialization of `invert_srgb_compand`. Moreover, although I don't clearly understand the reason, `RGB{Float64}`->`XYZ` seems to have no problem. (I guess the reason is the `@noinline`. `exp` potentially and definitely have the problem.)

In the case of [MWE](https://github.com/JuliaLang/julia/issues/35972#issue-622437194), `f()` will be eventually recompiled because `f()` is in a "shallow" place. However, in the Colors.jl, `cbrt` is in "deep" places from the public API (e.g. `colordiff` and `distinguishable_colors`), so it is difficult to avoid the problem.

Even if this problem is fixed by a compiler improvement, that improvement may not be backported to older versions. (It may not even be backported to v1.4 depending on the release timing of ~v1.4.2~ v1.4.3.) So I think we should look for the workarounds which can be handled in Colors.jl.

One simple workaround is not to precompile the functions which cause the problem. However, this detracts from the benefits of precompilation.

kimikage:
Although I don't want to do it much, specializing `cbrt` in Colors.jl is an option.  Since it is sufficient for Colors.jl to be able to calculate mainly for the [0, 1] domain, some optimizations may be available.

kimikage:
[...]

PR #427, opened 2020-05-27
Improve type stability in conversion methods

kimikage:
I found some type instability issues while tackling JuliaLang/julia#35972 or #425. Unfortunately, this doesn't solve JuliaLang/julia#35972, and doesn't provide much performance improvement.

~This is a successor of PR #424.~ (Merged. However, I'm still looking for a way to solve #425.)
 This should be the next v0.13 material since this changes the conversion results slightly.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -320,20 +320,25 @@ function cnvt(::Type{XYZ{T}}, c::xyY) where T
 end
 
 
-const xyz_epsilon = 216 / 24389
-const xyz_kappa   = 24389 / 27
+const xyz_epsilon   = 216 / 24389 # (6/29)^3
+const xyz_kappa     = 24389 / 27  # (29/6)^3*8
+const xyz_kappa_inv = 27 / 24389
 
 function cnvt(::Type{XYZ{T}}, c::Lab, wp::XYZ = WP_DEFAULT) where T
     fy = (c.l + 16) / 116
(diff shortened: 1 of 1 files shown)


src/conversions.jl
```diff
@@ -316,20 +316,25 @@ function cnvt(::Type{XYZ{T}}, c::xyY) where T
...
-const xyz_kappa   = 24389 / 27
+const xyz_epsilon   = 216 / 24389 # (6/29)^3
+const xyz_kappa     = 24389 / 27  # (29/6)^3*8
+const xyz_kappa_inv = 27 / 24389
 
 function cnvt(::Type{XYZ{T}}, c::Lab, wp::XYZ = WP_DEFAULT) where T
     fy = (c.l + 16) / 116
-    fx = c.a / 500 + fy
+    fx = fy + c.a / 500
     fz = fy - c.b / 200
```
kimikage, line 326: If we change these divisions into reciprocal multiplications, this method can be inlined, i.e. successfully vectorized. However, even if only `Lab`-->`XYZ` conversion is speeded up, `Lab`-->`RGB` conversion is not speeded up.

kimikage:
I separated part of this PR to #442.

kimikage:
Now that we have added the support for ColorTypes v0.11, we will start developing Colors v0.13 series. This means that breaking changes can be merged.

PR #429, opened 2020-06-07
Fix problem with `hex(::Gray{Bool})`

### LINKED ISSUE #428: hex() doesn't support Gray{Bool}
I noticed this because Juno throws an error at me when it tries to display a Gray{Bool} pixel. I'm trying to understand the hex() code a bit more before attempting a fix, but I assume this shouldn't be too complicated.

kimikage:
Fixes #428


kimikage:
Thank you for merging and releasing.

PR #433, opened 2020-07-04
[CI] switch from Travis to Github Action

johnnychen94:
Changes:

* remove appveyor (and also windows 32bit)
* ~~leave travis for Documentation and Coverage~~ coverage report and documentation are moved to Github Action, too.
* caching artifacts dir in CI


johnnychen94:
AppVeyor, however, needs to be uninstalled after this. Otherwise, it will give a big ❌ in the CI status.

.github/workflows/Documentations.yml
```diff
@@ -0,0 +1,40 @@
...
+      - 'release-'
+    tags: '*'
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: ["1"]
+        julia-arch: [x86]
```
kimikage, line 17: What is the reason to specify x86 instead of x64? I think the documentation should be generated in a ”typical” environment.
More importantly, this is not referenced in `julia-actions/setup-julia@latest`.
https://github.com/JuliaGraphics/Colors.jl/pull/433/checks?check_run_id=876228283#step:3:4

I can't decide which is better, Julia v1.0 (LTS) or latest v1.

johnnychen94: Nice catch. This is originally copied from https://github.com/fredrikekre/Literate.jl/blob/master/.github/workflows/docs.yml and I believe it's an oversight here.

I've changed it to x64.

kimikage:
Thanks!

We already have multiple commits on gh-pages, so the warning below seems to be ignorable.
https://github.com/JuliaGraphics/Colors.jl/runs/884894599?check_suite_focus=true#step:6:21

johnnychen94:
FYI, to tell TagBot not include issues/PRs in release summary, you can tag them with `no changelog`

> https://github.com/JuliaRegistries/TagBot
> By default, the following labels are ignored:
>
> * changelog skip
> * duplicate
> * exclude from changelog
> * invalid
> * no changelog
> * question
> * wont fix

kimikage:
Personally I don't feel the need to exclude this, though.:smile:

johnnychen94:
Heads up: TagBot needs a ssh key to trigger the docs CI when there's a new release.

Check the "TagBot & tagged versions" warning for more details: https://juliadocs.github.io/Documenter.jl/dev/man/hosting/#GitHub-Actions

PR #436, opened 2020-08-19
Improve precompilation with concrete typing

timholy:
`MSC` was being called during precompilation with non-inferrable argument types and consequently was an invalidation tigger (and risk). This also specifies concrete types for some of our bigger functions, which makes them more likely to compile only once, and then uses small normalization wrappers to handle argument diversity.

This cuts out more than a dozen invalidations.


src/parse.jl
```diff
@@ -150,6 +150,7 @@ function _parse_colorant(desc::AbstractString)
 
     error("Unknown color: ", desc)
 end
+_parse_colorant(desc::AbstractString) = _parse_colorant(String(desc))
```
kimikage, line 153: The `_parse_colorant` is an internal function, so I don't think we need to provide this method. Alternatively, we can force the conversion at higher levels (i.e. `Base.parse`).
```julia
Base.parse(::Type{C}, desc::AbstractString) where {C<:Colorant} = _parse_colorant(C, supertype(C), String(desc))
```

kimikage: Also, I think we can rewrite the workaround as follows in Julia v1:
```julia
function Base.parse(::Type{C}, desc::AbstractString) where {C<:Colorant}
    c = _parse_colorant(String(desc))::Colorant # BTW, we know that `c` is a `Colorant` 
    supertype(C) <: Colorant ? convert(C, c)::C : c
end
```

src/colormaps.jl
```diff
@@ -295,8 +295,10 @@ function colormap(cname::AbstractString, N::Int=100; mid=0.5, logscale=false, kv
             end
         end
 
-        return sequential_palette(p[1], N, w=p[2], d=p[3], c=p[4], s=p[5], b=p[6], wcolor=p[7],
-                                  dcolor=p[8], logscale=logscale)
+        # To avoid invalidation risk, it's best to make the call in a manner that inference knows the types
+        return sequential_palette(p[1]::Int, N, w=p[2]::Float64, d=p[3]::Float64, c=p[4]::Float64,
+                                  s=p[5]::Float64, b=p[6]::Float64, wcolor=p[7]::RGB{N0f8},
+                                  dcolor=p[8]::RGB{N0f8}, logscale=logscale)
```
kimikage, line 301: I think these are overly specific. The acceptable types are not documented. IMO, constraining the colors to `RGB{N0f8}` is problematic.

kimikage:
> This cuts out more than a dozen invalidations.

How was that number measured? I think the value is highly dependent on the packages you load.

The reason I have a concern with this is because of its relevance to PR #427.  If necessary, I can split the PR #427 into fixes related to the `MSC` and others.

As mentioned in https://github.com/JuliaGraphics/ColorTypes.jl/pull/213#issuecomment-673599021, I'm beware of breaking changes in Colors.jl.

src/parse.jl
timholy: Indeed I don't think we need the `supertype` stuff at all now.

timholy:
> How was that number measured? I think the value is highly dependent on the packages you load.

Just `using Colors` in a fresh Julia session (except for having loaded SnoopCompileCore). The number of invalidations can be counted using `length(uninvalidated(invs))` from SnoopCompile. And yes, the number depends very much on what else you load, at least until the community starts tackling these issues more broadly.

timholy:
> If necessary, I can split the PR #427 into fixes related to the MSC and others.

I don't think that's necessary, this is quite orthogonal to that work. Which, by the way, is very welcome in its own right.

src/parse.jl
kimikage: That's right. However, I am concerned that `convert` will result in a slight degradation in speed performance. (ColorTypes.jl defines the method to [pass through](https://github.com/JuliaGraphics/ColorTypes.jl/blob/6d2759c6becc5ab282a0bee6c257b179d7e63ed8/src/conversions.jl#L72) and I'm not sure why it affects the speed. :confused:)

Even so, it does improve the following strange behavior.:smile:
```julia
julia> parse(Colorant{Float32}, "red")
RGB{N0f8}(1.0,0.0,0.0) # current
RGB{Float32}(1.0f0,0.0f0,0.0f0) # force conversion
```

kimikage:
I didn't see a clear difference on the nightly build (Commit bd65e85ca8) for Windows, and I don't know about `uninvalidated()`.

However, the revised changes in this PR seem to make things better.

src/parse.jl
timholy: The optimizer elides the call to `convert` when it's a no-op:

```julia
julia> @code_typed parse(Colorant, "red")
CodeInfo(
1 ─ %1 = invoke Colors._parse_colorant(_3::String)::Any
│        Core.typeassert(%1, Colors.Colorant)::Colorant
│   %3 = π (%1, Colorant)
└──      return %3
) => Colorant

julia> @code_typed parse(RGB{N0f8}, "red")
CodeInfo(
1 ─ %1 = $(Expr(:static_parameter, 1))::Core.Const(RGB{N0f8}, false)
│   %2 = invoke Colors._parse_colorant(_3::String)::Any
│        Core.typeassert(%2, Colors.Colorant)::Colorant
│   %4 = π (%2, Colorant)
│   %5 = Colors.convert(%1, %4)::Any
│        Core.typeassert(%5, $(Expr(:static_parameter, 1)))::RGB{N0f8}
│   %7 = π (%5, RGB{N0f8})
└──      return %7
) => RGB{N0f8}
```
The type-assert is necessary because `_parse_colorant` has a sufficiently-complicated `Union` that inference justifiably chooses to declare it `Any`.

kimikage: I can't find anything that causes the slowdown as you've shown, but the new implementation slightly slows it down in the benchmarks. The `code_native` output of `parse` is so long that I don't feel like analyzing it right now.

```julia
julia> julia> versioninfo()
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)

julia> @btime parse(Colorant, "red");
  605.747 ns (3 allocations: 128 bytes) # before
  645.455 ns (3 allocations: 128 bytes) # after
```

```julia
julia> versioninfo()
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)

julia> @btime parse(Colorant, "red");
  335.023 ns (3 allocations: 128 bytes) # before
  373.659 ns (3 allocations: 128 bytes) # after
```

timholy: Hmm, that is surprising. It doesn't seem to be fixed by the `supertype` check. I am a bit puzzled. I'd stilll say this is an improvement, and that we can look into the optimizer issues later.

timholy:
OK to merge?

src/parse.jl
kimikage: I added a `::Colorant` assertion, thinking it would rather help the optimizer, but it seems to be the cause.:sob:

timholy: You probably saw this, but for the benefit of all: xref https://github.com/JuliaLang/julia/issues/37135.

timholy: Check out this implementation:
```julia
function Base.parse(::Type{C}, desc::AbstractString) where {C<:Colorant}
    c = _parse_colorant(String(desc))
    return isconcretetype(C) ? C(c)::C : c
end

julia> @btime parse(Colorant, "red");
  404.800 ns (5 allocations: 144 bytes)

julia> @btime parse(RGB{N0f8}, "red");
  501.647 ns (5 allocations: 144 bytes)
```

Yet of course the second will be better if the return type is used in other code, since it will assure the compiler that it can do compile-time lookup of all the calls.

I vote we go with this implementation?

timholy: Hmm, that one was wrong...how about this:
```julia
function Base.parse(::Type{C}, desc::AbstractString) where {C<:Colorant}
    c = _parse_colorant(String(desc))
    return C === Colorant    ? c :
           isconcretetype(C) ? convert(C, c)::C : convert(C, c)
end

julia> @btime parse(Colorant, "red");
  407.970 ns (5 allocations: 144 bytes)

julia> @btime parse(RGB{N0f8}, "red");
  539.243 ns (5 allocations: 144 bytes)

julia> @btime parse(RGB, "red");
  521.741 ns (5 allocations: 144 bytes)
```

kimikage:
It might be a good practice to check the performance with `SubString` input, just to be sure. I'm writing from my phone right now and haven't tested it.

timholy:
FYI https://github.com/JuliaLang/julia/pull/37192 gave me this:
```julia
julia> @btime parse(Colorant, "red");
  341.559 ns (1 allocation: 16 bytes)
```
compared to
```julia
julia> @btime parse(Colorant, "red");
  407.970 ns (5 allocations: 144 bytes)
```
above.

PR #439, opened 2020-08-24
Add error message for `colordiff` with transparent color inputs

### LINKED ISSUE #437: colordiff not work with all colors?
Just wondering why `colordiff` doesn't work with all colors?

```
julia> using Colors
julia> colordiff(RGBA(0.35, 0.18, 0.90, 1), RGBA(0.34, 0.18, 0.90, 1))
ERROR: MethodError: no method matching colordiff(::RGBA{Float64}, ::RGBA{Float64})
Stacktrace:
 [1] top-level scope at none:1
```

There's probably a good reason, but I'm interested to know what it is. 

Colors v0.12.3, Julia 1.5.0

kimikage:
This adds the support for `TransparentColor` inputs of `colordiff`. If the alpha of the `TransparentColor` input is not 1, an error is thrown. (There is room for improvement in the content of the message.)
```julia
julia> colordiff(RGBA(0.35, 0.18, 0.90, 1), RGBA(0.34, 0.18, 0.90, 1))
0.2965835193560762

julia> colordiff(RGBA(0.35, 0.18, 0.90, 0.5), RGBA(0.34, 0.18, 0.90, 0.5))
ERROR: ArgumentError: cannot evaluate the difference in transparent colors.
  Their appearance depends on the backdrop.
```
This also adds basic tests for `colordiff` and fixes the conversion error within `DE_BFD`.

Fixes #437

### DIFF
--- a/test/colordiff.jl
+++ b/test/colordiff.jl
@@ -1,4 +1,4 @@
-using Colors
+using Colors, Test
 
 # Test the colordiff function against example input and output from:
 #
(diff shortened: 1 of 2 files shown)


test/colordiff.jl
```diff
@@ -47,16 +47,56 @@ using Colors
...
-    a, b = rand(100), rand(100)
-    @test all(@. colordiff(a, b) == colordiff(Gray(a), b) == colordiff(a, Gray(b)) == colordiff(Gray(a), Gray(b)))
+    jl_red    =    RGB{N0f8}(Colors.JULIA_LOGO_COLORS.red)
+    jl_green  = RGB{Float32}(Colors.JULIA_LOGO_COLORS.green)
+    jl_blue   = HSV{Float64}(Colors.JULIA_LOGO_COLORS.blue)
+    jl_purple = Lab{Float32}(Colors.JULIA_LOGO_COLORS.purple)
+    colors = (jl_red, jl_green, jl_blue, jl_purple)
+    pairs = ((a, b) for a in colors for b in Iterators.filter(c -> c != a,colors))
+
+    @testset "properties of metrics" begin
```
timholy, line 65: Love it!

kimikage:
This PR does not change `distinguishable_colors`. If you think of a good error message, you can suggest it.

PR #440, opened 2020-08-25
Generalize XYZ and LMS arithmetic

### LINKED ISSUE #435: promotion of types for `+`/`-` on linear color spaces with mixed underlying representation
Definitions at https://github.com/JuliaGraphics/Colors.jl/blob/0e37ed40acd30de13560fe904f4b73ad9cef94f7/src/algorithms.jl#L2-L5

don't allow

```julia
julia> XYZ{Float32}(0,0,0) + XYZ{Float64}(0,0,0)
ERROR: MethodError: no method matching +(::XYZ{Float32}, ::XYZ{Float64})
```

kimikage:
Fixes #435

I think this could be v0.12 material.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1,11 +1,12 @@
 # Arithmetic
 #XYZ and LMS are linear vector spaces
 const Linear3 = Union{XYZ, LMS}
-+(a::C, b::C) where {C<:Linear3} = C(comp1(a)+comp1(b), comp2(a)+comp2(b), comp3(a)+comp3(b))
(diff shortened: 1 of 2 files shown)


kimikage:
For arithmetic for the colors defined in ColorTypes.jl (e.g. `XYZ` and `LMS`), either ColorTypes.jl, Colors.jl or ColorVectorSpace.jl is the appropriate place to define the arithmetic. In the unlikely event that this change conflicts with your package, please report the issue.

I'm going to leave the definition of the identities (e.g. `oneunit`) to ColorTypes.jl.

PR #444, opened 2020-09-19
Add ssh key for TagBot.yml

### LINKED ISSUE #443: configure SSH key for TagBot and optionally for CompatHelper
I tried to not notify anyone with a direct commit c88f062 but it doesn't work...

Here's the motivation: 

One GitHub action could not trigger another GitHub action if only configured with `token: ${{ secrets.GITHUB_TOKEN }}`. This limitation means, PR created by CompatHelper could not trigger the unit test, and release/tag created by tagbot could trigger the docs CI, which is switched to GH in #433 .

The ssh key for tagbot seems a necessary step. Though the ssh key for compat helper could be skipped given that we still need to manually check the versions even if the unit test runs.

For TagBot, the quick command to generate ssh key:

```
ssh-keygen -t rsa -b 4096 -C "email@example.com" -f tagbot
```

then copy `tagbot.pub` as keys and `tagbot` as secrets (let's say with name "TagBot"). A commit like https://github.com/JuliaImages/ImageEdgeDetection.jl/commit/bd6515ae2bb6a88842cc07fd6abf1ea40231d6bd should then set things done. (Haven't checked it because there's no new releases...)

References:

- https://github.com/JuliaRegistries/CompatHelper.jl#12-set-up-the-ssh-deploy-key-optional
- https://juliadocs.github.io/Documenter.jl/dev/man/hosting/#GitHub-Actions

cc: @cormullion @timholy

timholy:
Closes #443

Hopefully I did all the steps correctly...if not, let me know.

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -9,3 +9,4 @@ jobs:
       - uses: JuliaRegistries/TagBot@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          ssh: ${{ secrets.TagBot }}


johnnychen94 (approved):
We wouldn't know until there's a new release 😆

.github/workflows/TagBot.yml
```diff
@@ -9,3 +9,4 @@ jobs:
       - uses: JuliaRegistries/TagBot@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          ssh: ${{ secrets.TagBot }}
```
johnnychen94, line 12: It probably should be `secrets.TAGBOT` because GitHub seems to force uppercase for all its secrets. Unsure whether this version works.

PR #451, opened 2020-11-23
fix type instability in weighted_color_mean

rafaqz:
Removes the use of `convert` inside `mapc` which for some reason isn't type-stable. It isn't needed anyway as the previous conversion of `weighted1` has the same effect.

~~This type instability doesn't have a major performance impact, but makes it difficult to test inference in other packages that use this where type instability does have a major impact.~~

See JuliaGraphics/ColorSchemes.jl#52

### DIFF
--- a/test/utilities.jl
+++ b/test/utilities.jl
@@ -186,6 +186,7 @@ using Colors, FixedPointNumbers, Test, InteractiveUtils
                 C == AGray32 && (T=N0f8)
                 c1 = C(T(0),T(1))
                 c2 = C(T(1),T(0))
+                @inferred weighted_color_mean(0.5,c1,c2)
                 @test weighted_color_mean(0.5,c1,c2) == C(T(1)-T(0.5),T(0.5))
             end
         end
@@ -195,11 +196,13 @@ using Colors, FixedPointNumbers, Test, InteractiveUtils
                 if C<:Color
                         c1 = C(T(1),T(1),T(0))
                         c2 = C(T(0),T(1),T(1))
+                    @inferred weighted_color_mean(0.5,c1,c2)
                     @test weighted_color_mean(0.5,c1,c2) == C(T(0.5),T(0.5)+T(1)-T(0.5),T(1)-T(0.5))
                 else
                         C == ARGB32 && (T=N0f8)
                         c1 = C(T(1),T(1),T(0),T(1))
                         c2 = C(T(0),T(1),T(1),T(0))
+                    @inferred weighted_color_mean(0.5,c1,c2)
                     @test weighted_color_mean(0.5,c1,c2) == C(T(0.5),T(0.5)+T(1)-T(0.5),T(1)-T(0.5),T(0.5))
                 end
             end
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -203,12 +203,11 @@ function weighted_color_mean(w1::Real, c1::C, c2::C) where {Cb <: Union{HSV, HSL
                                                             C <: Union{Cb, AlphaColor{Cb}, ColorAlpha{Cb}}}
     normalize_hue(_weighted_color_mean(w1, c1, c2))
 end
-function _weighted_color_mean(w1::Real, c1::Colorant, c2::Colorant)
+function _weighted_color_mean(w1::Real, c1::Colorant{T1}, c2::Colorant{T2}) where {T1,T2}
     @fastmath min(w1, oneunit(w1) - w1) >= zero(w1) || throw(DomainError(w1, "`w1` must be in [0, 1]"))
-    T = promote_type(eltype(c1), eltype(c2))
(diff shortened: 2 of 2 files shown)


rafaqz:
Don't merge this. It passes tests here but breaks other packages. I'll try to resolve.

rafaqz:
Ok. Failures in ColorSchemes.jl are just minor floating-point errors from broadcast vs non-broadcast versions.

Probably the new-found type stability is letting the compiler reorganize something somewhere that it couldn't before.

timholy:
Do you have an example of inputs for which the current version isn't inferrable? I'm surprised it's failing, might merit a deeper investigation.

rafaqz:
```julia
using Colors, Test
julia> @inferred weighted_color_mean(0.5, RGB(0.9, 0.6, 0.1), RGB(0.9, 0.7, 0.2))
ERROR: return type RGB{Float64} does not match inferred return type RGB
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] top-level scope at REPL[5]:1

julia> @code_warntype weighted_color_mean(0.5, RGB(0.9, 0.6, 0.1), RGB(0.9, 0.7, 0.2))
Variables
  #self#::Core.Compiler.Const(Colors.weighted_color_mean, false)
  w1::Float64
  c1::RGB{Float64}
  c2::RGB{Float64}

Body::RGB
1 ─ %1 = Colors._weighted_color_mean(w1, c1, c2)::RGB
└──      return %1

```

rafaqz:
BTW this is on julia v"1.5.2"

kimikage:
I don't think there is anything wrong with this PR. However, in the next minor release I plan to change the behavior. (cf. PR #441)
This means that the problem may recur sooner or later. So, I think we need to clarify the causes of failure in inference.

kimikage:
I'm not sure of the reason, but the inference seems to work if we move the `promote_type` into the anonymous function.

PR #441 version:
```julia
function _weighted_color_mean(w1::Real, c1::Colorant{T1}, c2::Colorant{T2}) where {T1, T2}
    @fastmath min(w1, oneunit(w1) - w1) >= zero(w1) || throw(DomainError(w1, "`w1` must be in [0, 1]"))
    w2 = oneunit(w1) - w1
    mapc((x, y)->convert(promote_type(T1, T2), muladd(w1, x, w2 * y)), c1, c2)
end
```
Apart from readability, you can also replace it with `promote_type(typeof(x), typeof(y))`.

It's interesting, but still weird. :confused:

rafaqz:
`mapc` [isn't inlined](https://github.com/JuliaGraphics/ColorTypes.jl/blob/master/src/operations.jl#L108) in ColorTypes.jl. but changing that didn't fix it either for me, and neither did using a let block.

Maybe type information for the `DataType` `T` is lost when passed into the closure? I guess the field type of `T` in the closure struct would just be `DataType`... so maybe it has to all inline and compile away completely to actually be type-stable.

I like the solution of moving `promote_type` inside the closure. Easy.

rafaqz:
Anyway, lets go with your version, and close this. When do you intend to make a minor release? I'm also keen to lock down inference in ColorSchemes.jl whenever this is merged.

kimikage:
I think we can also keep the current numerical behavior and only improve the inferrence. It should be able to be released as a patch version.

Perhaps the next minor release trigger is in https://github.com/JuliaGraphics/ColorTypes.jl/pull/213.

rafaqz:
Ok I ended up with a hybrid version where convert is done twice. Otherwise there are small rounding errors in the tests.

```julia
function _weighted_color_mean(w1::Real, c1::Colorant{T1}, c2::Colorant{T2}) where {T1,T2}
    @fastmath min(w1, oneunit(w1) - w1) >= zero(w1) || throw(DomainError(w1, "`w1` must be in [0, 1]"))
    weight1 = convert(promote_type(T1, T2), w1) # TODO: Consider the need for this
    weight2 = oneunit(weight1) - weight1
    mapc((x, y) -> convert(promote_type(T1, T2), muladd(weight1, x, weight2 * y)), c1, c2)
end
```

test/utilities.jl
```diff
@@ -176,6 +176,7 @@ using Colors, FixedPointNumbers, Test, InteractiveUtils
         gray_b1 = Gray{Bool}(1)
         gray_b0 = Gray{Bool}(0)
         @test weighted_color_mean(0, gray_b1, gray_b0) === gray_b0
+        @test weighted_color_mean(0, gray_b1, gray_b0) === gray_b0
```
kimikage, line 179: Did you mean `@inferred`?

rafaqz: Yeah, but I ended up putting inferred in the `colorElementTypes` loop. Removed.

kimikage (approved):
Thank you for catching this problem and fixing it.

After a bit of time I will merge this and make a patch release.

PR #449, opened 2020-11-06
Update TagBot.yml

kimikage:
cf. https://discourse.julialang.org/t/ann-required-updates-to-tagbot-yml/49249

// Sorry for the messy branch name as I thought I had created the branch in my fork.

PR #452, opened 2020-12-01
Version 0.12.5

kimikage:
TagBot must be updated before registering v0.12.5 (cf. #449).


johnnychen94:
I believe the old tagbot is still functioning? ImageCore v0.8.19 for example.

kimikage:
Perhaps that is right. However, that will postpone the confirmation of updating and working well.

PR #461, opened 2021-03-02
Fix `DE_94` & `DE_CMC` color difference metrics

### LINKED ISSUE #438: Non-standard implementation of `DE_CMC` and `DE_94`
The CMC(l:c) color difference model quantifies the difference from a standard (reference) color. 
https://en.wikipedia.org/wiki/Color_difference#CMC_l:c_(1984)

However, the current implementation uses the mean value as the "standard" rather than one of the `colordiff` inputs.
https://github.com/JuliaGraphics/Colors.jl/blob/efdb8a569e1607517e986fa01a51f6e41ab9cbd0/src/differences.jl#L328-L329
The comment implies that it is intentional rather than a mistake, but the averaging is not the original method. Even if we were to follow the method of using the mean values, the following uses `a.l` instead of `ml`, which is not consistent.
https://github.com/JuliaGraphics/Colors.jl/blob/efdb8a569e1607517e986fa01a51f6e41ab9cbd0/src/differences.jl#L338-L343

Although CMC (l:c) is an old model, it appears to still be used in the textile industry, as it is included in the ISO 105-J03 standard.

In practice, there seems to be a method of substituting a reference with an average color of multiple samples of the same product when no reference is available, but that is not the same as averaging inside `colordiff`.

However, this fix is too drastic to be at the patch level, so I would like to fix this in v0.13.0.

kimikage:
This makes `DE_94` and `DE_CMC` quasimetric. Within those metrics, the first argument of `colordiff` is taken as the reference color. This adds the weight parameters `k1` and `k2` to `DE_94`.

This also makes the omission of the metric constructor arguments more flexible.

~**Edit:** `DE_94` seems to have a similar problem. I will fix it as well.~ Done

Fixes #438

### DIFF
--- a/src/differences.jl
+++ b/src/differences.jl
@@ -10,34 +10,52 @@ struct DE_2000 <: DifferenceMetric
     kl::Float64
     kc::Float64
     kh::Float64
+    DE_2000(kl=1, kc=1, kh=1) = new(kl, kc, kh)
 end
 """
-    DE_2000()
-    DE_2000(kl, kc, kh)
+    DE_2000(kl=1, kc=1, kh=1)
 
 Construct a metric of the CIE Delta E 2000 recommendation, with weighting
 parameters `kl`, `kc` and `kh` as provided for in the recommendation. When not
 provided, these parameters default to `1`.
 """
-DE_2000() = DE_2000(1,1,1)
+DE_2000()
 
 
 struct DE_94 <: DifferenceMetric
     kl::Float64
     kc::Float64
     kh::Float64
+    k1::Float64
+    k2::Float64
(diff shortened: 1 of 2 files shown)


src/differences.jl
```diff
@@ -365,7 +371,7 @@ end
 function _colordiff(ai::Color, bi::Color, m::DE_BFD)
     # Currently, support for the `wp` argument of `convert` is limited.
     function to_xyz(c::Color, wp)
-        c isa XYZ && return ai
+        c isa XYZ && return c
```
kimikage, line 374: :sob:

I am not going to backport this to v0.12. As a workaround, if the second argument is `XYZ`, please convert it to something other than `XYZ` (e.g. `xyY`).

PR #463, opened 2021-03-03
Make display of color swatches slightly robust

kimikage:
This reduces the risk of having invalid suffixes when converting numbers to strings. This also supports indices that start with other than 1.
There is almost no improvement in performance.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -71,9 +71,10 @@ function Base.show(io::IO, mime::MIME"image/svg+xml",cs::AbstractVector{T};
         </defs>
         <rect width="$n" height="1" fill="url(#$id)" />
         """)
-    for i in 1:min(n, actual_max_swatches)
-        hexc = hex(color(cs[i]))
-        opacity = string(round(alpha(cs[i]), digits=4))
+    for (i, c) in enumerate(cs)
+        i > actual_max_swatches && break
+        hexc = hex(color(c))
+        opacity = string(round(float(alpha(c)), digits=4))
         op = replace(opacity, r"(^0(?!\.0$))|(\.0$)"=>"")
         write(io,
             """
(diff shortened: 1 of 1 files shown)


src/display.jl
```diff
@@ -71,9 +71,10 @@ function Base.show(io::IO, mime::MIME"image/svg+xml",cs::AbstractVector{T};
         </defs>
         <rect width="$n" height="1" fill="url(#$id)" />
         """)
-    for i in 1:min(n, actual_max_swatches)
-        hexc = hex(color(cs[i]))
-        opacity = string(round(alpha(cs[i]), digits=4))
+    for (i, c) in enumerate(cs)
+        i > actual_max_swatches && break
+        hexc = hex(color(c))
+        opacity = string(round(float(alpha(c)), digits=4))
```
kimikage, line 77: `alpha()` may return a `FixedPoint` number. FixedPointNumbers v0.8 does not support `round(x, digits=n)`, and the fallback will return floating-point numbers. However, this problem has been fixed in https://github.com/JuliaMath/FixedPointNumbers.jl/pull/235.
```julia
julia> round(0.5N0f8, digits=4)
0.502f0 # FPN v0.8
0.502N0f8 # FPN #master
```
Here is another problem. Unlike `Float32` etc., `string(::FixedPoint)` currently returns a string with the suffix.
```julia
julia> string(0.5N0f8)
"0.502N0f8"

julia> string(0.5f0)
"0.5"
```
Perhaps `string(::FixedPoint)` should return a string without suffix, but that is a matter of FixedPointNumbers.

mcabbott:
Would it be possible to make a release with this PR? Some weird errors in Pluto and some `git blame` eventually led me to discover that the problem has been fixed, I think.

kimikage:
Colors v0.12.7 was released. If the issue is still not resolved, please report it.

mcabbott:
Thanks!

PR #466, opened 2021-03-22
Fix stack overflow in undefined color space conversions

### LINKED ISSUE #465: `StackOverflowError` when attempting undefined color space conversion
Although undefined color space conversion should result in an error, the error message is not helpful.
```julia
julia> using ColorTypes

julia> struct MyColor{T} <: Color{T, 3}; c1::T; c2::T; c3::T; end

julia> convert(RGB, MyColor{Float32}(1, 2, 3)) # without Colors.jl
ERROR: No conversion of MyColor{Float32}(1.0f0,2.0f0,3.0f0) to RGB{Float32} has been defined

julia> using Colors

julia> convert(RGB, MyColor{Float32}(1, 2, 3)) # with Colors.jl
ERROR: StackOverflowError:
```

A fix for this issue could theoretically affect downstream packages. However, given that there have been no issue reports on this so far, I think it can be included in the v0.12 series as a bug fix in practice.

kimikage:
This avoids stack overflow in conversions for color types which are not defined in ColorTypes.jl.

This also attempts to convert via `RGB` instead of directly to the destination type as a fallback. That improves support for source types other than `Color3`. In other words, if you define a new color type, you should be able to achieve conversions with other color types, at least by implementing conversions with `(Abstract)RGB`.

Technically speaking, this is a breaking change, but I don't think it will have any negative impact unless the conversion methods are overloaded improperly.

Fixes #465

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -50,7 +50,6 @@ end
 
 ColorTypes._convert(::Type{Cdest}, ::Type{Odest}, ::Type{Osrc}, c) where {Cdest<:Color,Odest,Osrc} = cnvt(Cdest, c)
 
-
 # with the whitepoint `wp` as the third argument
 convert(::Type{XYZ}, c, wp::XYZ) = cnvt(XYZ{eltype(wp)}, c, wp)
 convert(::Type{Lab}, c, wp::XYZ) = cnvt(Lab{eltype(wp)}, c, wp)
@@ -62,12 +61,13 @@ convert(::Type{Luv{T}}, c, wp::XYZ) where {T} = cnvt(Luv{T}, c, wp)
 
 # Fallback to catch undefined operations
 cnvt(::Type{C}, c::TransparentColor) where {C<:Color} = cnvt(C, color(c))
-cnvt(::Type{C}, c) where {C} = convert(C, c)
+cnvt(::Type{C}, c) where {C} = convert(C, convert(RGB{eltype(C)}, c)::RGB{eltype(C)})
 
 # Conversions from grayscale
 # --------------------------
-cnvt(::Type{C}, g::AbstractGray) where {C<:Color3}  = cnvt(C, convert(RGB{eltype(C)}, g))
-
+function ColorTypes._convert(::Type{Cdest}, ::Type{Odest}, ::Type{Osrc}, g) where {Cdest<:Color,Odest,Osrc<:AbstractGray}
+    cnvt(Cdest, convert(RGB{eltype(Cdest)}, g))
+end
 
 macro mul3x3(T, M, c1, c2, c3)
     esc(quote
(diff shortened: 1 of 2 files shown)


src/conversions.jl
```diff
@@ -62,12 +61,13 @@ convert(::Type{Luv{T}}, c, wp::XYZ) where {T} = cnvt(Luv{T}, c, wp)
 
 # Fallback to catch undefined operations
 cnvt(::Type{C}, c::TransparentColor) where {C<:Color} = cnvt(C, color(c))
-cnvt(::Type{C}, c) where {C} = convert(C, c)
+cnvt(::Type{C}, c) where {C} = convert(C, convert(RGB{eltype(C)}, c)::RGB{eltype(C)})
```
kimikage, line 64: As long as the chain `Base.convert`-->`ColorTypes.cconvert`-->`ColorTypes._convert`-->`Colors.cnvt` is followed, the fact that this method is called means that `convert(C, c)` will eventually call this method again, i.e. will cause stack overflow. (There may have once been a path where `cnvt` was called without calling `Base.convert`, but I'm not sure.)

Therefore, I think it is reasonable to try to convert via a different color space. The candidates for the transit point are `RGB`, `XYZ`, and `Lab`, but for practical purposes, I think `RGB` is the most appropriate.

src/conversions.jl
```diff
@@ -62,12 +61,13 @@ convert(::Type{Luv{T}}, c, wp::XYZ) where {T} = cnvt(Luv{T}, c, wp)
...
 
 # Conversions from grayscale
 # --------------------------
-cnvt(::Type{C}, g::AbstractGray) where {C<:Color3}  = cnvt(C, convert(RGB{eltype(C)}, g))
-
+function ColorTypes._convert(::Type{Cdest}, ::Type{Odest}, ::Type{Osrc}, g) where {Cdest<:Color,Odest,Osrc<:AbstractGray}
+    cnvt(Cdest, convert(RGB{eltype(Cdest)}, g))
+end
```
kimikage, line 70: This has a higher priority than the other `cnvt`s, so it should be promoted to `ColorTypes._convert`. The fallback for `cnvt` is via `RGB` as well, so we can continue to pass a gray to the second argument of `cnvt`.
```julia
julia> Colors.cnvt(LCHab{Float32}, Gray(1))
LCHab{Float32}(100.00001f0,2.6347714f-5,161.63281f0)
```

kimikage:
If there are no objections to this before we support ColorTypes v0.11, I would like to include this in the Colors v0.12 series.

kimikage:
ColorTypes v0.11 is now ready for release, so I will merge this.

kimikage:
Float16-related changes on nightly seem to cause the test failure. It should not be directly related to this PR.

PR #468, opened 2021-04-02
Version 0.12.7

kimikage:
`InteractiveUtils` is not used (imported) in `module Colors`, and is a standard library. Therefore, it should be safely removed from the `[deps]` section.

### DIFF
--- a/docs/src/colormapsandcolorscales.md
+++ b/docs/src/colormapsandcolorscales.md
@@ -23,7 +23,7 @@ julia> c2 = colorant"green"
 RGB{N0f8}(0.0,0.502,0.0)
 
 julia> range(c1, stop=c2, length=15)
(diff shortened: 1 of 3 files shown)


docs/src/colormapsandcolorscales.md
```diff
@@ -23,7 +23,7 @@ julia> c2 = colorant"green"
 RGB{N0f8}(0.0,0.502,0.0)
 
 julia> range(c1, stop=c2, length=15)
-15-element Array{RGB{N0f8},1} with eltype RGB{FixedPointNumbers.Normed{UInt8,8}}:
+15-element Array{RGB{N0f8},1} with eltype RGB{FixedPointNumbers.N0f8}:
```
kimikage, line 26: In Julia v1.6, the alias is displayed.

PR #470, opened 2021-04-10
Implement `showable` to check that all components are finite

### LINKED ISSUE #469: `NaN`s are not showable
`Gray(NaN)` and `RGB(NaN)` are not showable:

```julia
julia> Gray(NaN)
Error showing value of type Gray{Float64}:
ERROR: ArgumentError: element type N0f8 is an 8-bit type representing 256 values from 0.0 to 1.0,
  but the values (NaN, 1.0) do not lie within this range.
  See the READMEs for FixedPointNumbers and ColorTypes for more information.
```

The error message is a bit misleading, too.

ImageShow uses `ImageCore.clamp01nan` to handle this, I think we can perhaps move `clamp01` and `clamp01nan` to `Colors`.

https://github.com/JuliaImages/ImageCore.jl/blob/a8f3d2c52070d9c166e1657ba25e72c8df00c742/src/map.jl#L3-L39

kimikage:
This is useful to avoid conversion errors in advance.
Since `NaN`/`Inf` may be caused by bugs, implicit clamping is not performed. This is in contrast to clamping for out-of-gamut colors.

Closes #469.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -6,6 +6,15 @@ const max_height = 150
 const max_swatch_size = 25
 const default_max_swatches = 128 * 128
 
+_isfinite(c::Colorant) = mapreducec(isfinite, &, true, c)
+
+Base.showable(::MIME"image/svg+xml", c::Colorant) = _isfinite(c)
+# Note that ImageShow.jl overloads `showable` for `AbstractMatrix{<:Color}`
+function Base.showable(::MIME"image/svg+xml", cs::Union{AbstractVector{<:Colorant},
+                                                        AbstractMatrix{<:Color}})
+    all(_isfinite, cs)
+end
+
 function Base.show(io::IO, mime::MIME"image/svg+xml", c::Color)
     write_declaration(io, mime)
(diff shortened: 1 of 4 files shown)


src/display.jl
```diff
@@ -6,6 +6,15 @@ const max_height = 150
 const max_swatch_size = 25
 const default_max_swatches = 128 * 128
 
+_isfinite(c::Colorant) = mapreducec(isfinite, &, true, c)
+
+Base.showable(::MIME"image/svg+xml", c::Colorant) = _isfinite(c)
+# Note that ImageShow.jl overloads `showable` for `AbstractMatrix{<:Color}`
```
johnnychen94, line 12:
```suggestion
# Note that ImageShow.jl overloads `showable` for `MIME"image/png"` and `AbstractMatrix{<:Colorant}`
```

johnnychen94 (approved):

kimikage:
If this is urgent, you can release v0.12.8.

PR #476, opened 2021-05-14
Clean up "src/differences.jl"

kimikage:
This exports the re-implemented version of `mean_hue`.
This is a preparation for future performance improvements, and is not in itself a change intended to improve accuracy or speed. This will slightly affect the calculation results.


kimikage:
Related to issue #477, the implementation of `color_diff` will be reworked.
If you find a bug in this PR, please report it.

PR #486, opened 2021-06-16
Improve `Lab`<-->`XYZ` conversions

kimikage:
This adds the separate methods for omitting the white points to facilitate constant folding.
This also suppresses the promotion of intermediate variables to `Float64` due to white point being `XYZ{Float64}`. (cf. issue #477)

A part of this PR has been separated from PR #482.

I'm not sure of the reason, but I avoid `mapc` for now because it makes the precompilation problem worse. This PR does not solve the precompilation problem. :confused:

PR #491, opened 2021-06-19
Improve `Luv`<-->`XYZ` conversions

### LINKED ISSUE #425: Consideration of workarounds for `cbrt` and `exp` precompilation problems

kimikage:
This is an `Luv` variation of the PR #486.

Closes #425

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -367,28 +367,31 @@ end
 
 
 function xyz_to_uv(c::XYZ)
+    F = typeof(4c.x / oneunit(c.x))
     d = c.x + 15c.y + 3c.z
-    d==0 && return (d, d)
-    u = 4c.x / d
-    v = 9c.y / d
-    return (u, v)
+    d == zero(d) && return (zero(F), zero(F))
+    return (4c.x, 9c.y) ./ d # (u, v)
 end
 
-
-function cnvt(::Type{XYZ{T}}, c::Luv, wp::XYZ = WP_DEFAULT) where T
+@inline function luv2xyz(::Type{XYZ{T}}, c, y_wp::F, uv_wp::Tuple{F, F}) where {T, F}
     c.l == zero(c.l) && return XYZ{T}(zero(T), zero(T), zero(T))
-
-    u_wp, v_wp = xyz_to_uv(wp)
-
-    ls = c.l * oftype(u_wp, xyz_kappa_inv)
-    y = c.l > 8 ? wp.y * ((c.l + 16) / 116)^3 : wp.y * ls
-
-    u = c.u / (13c.l) + u_wp
-    v = c.v / (13c.l) + v_wp
-    v4 = 4v
-    x = y * 9u / v4
-    z = y * (12 - 3u - 20v) / v4
+    ls = c.l * F(xyz_kappa_inv)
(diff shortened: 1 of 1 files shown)


kimikage:
### Benchmark
```julia
using Colors, BenchmarkTools
rgb_f64 = rand(RGB{Float64}, 1000, 1000);
xyz_f64 = XYZ{Float64}.(rgb_f64);
xyz_f32 = XYZ{Float32}.(xyz_f64);
luv_f64 = Luv{Float64}.(xyz_f64);
luv_f32 = Luv{Float32}.(luv_f64);
```
```julia
julia> @btime convert.(XYZ, $luv_f64);
  10.431 ms (2 allocations: 22.89 MiB) # before
  9.759 ms (2 allocations: 22.89 MiB)  # after

julia> @btime convert.(XYZ, $luv_f32);
  10.223 ms (2 allocations: 11.44 MiB) # before
  8.600 ms (2 allocations: 11.44 MiB)  # after

julia> @btime convert.(Luv, $xyz_f64);
  19.246 ms (2 allocations: 22.89 MiB) # before (recompiled)
  15.082 ms (2 allocations: 22.89 MiB) # after

julia> @btime convert.(Luv, $xyz_f32);
  20.847 ms (2 allocations: 11.44 MiB) # before
  14.830 ms (2 allocations: 11.44 MiB) # after
```

PR #482, opened 2021-06-02
Use `Float32` for the matrix coefficients for low precision colors

kimikage:
This also changes the definition of the sRGB transform matrix slightly. (cf. https://github.com/JuliaGraphics/Colors.jl/issues/355#issuecomment-539524932)

cf. #477

PR #493, opened 2021-06-20
Optimize `pow5_12(::Float32)`

kimikage:
Without PR #482, `pow5_12(::Float32)` was never called because linear RGB values are promoted to `Float64` in the transform matrix multiplication.
The new `pow5_12(::Float32)` has an accuracy of about 0.5 ULP.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -81,8 +81,9 @@ correct_gamut(c::CV) where {CV<:TransparentRGB} =
 @inline function srgb_compand(v)
     F = typeof(0.5f0v) === Float32 ? Float32 : promote_type(Float64, typeof(v))
     vf = F(v)
+    vc = @fastmath max(vf, F(0.0031308))
     # `pow5_12` is an optimized function to get `v^(1/2.4)`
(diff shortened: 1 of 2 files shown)


kimikage:
### Benchmark
cf. https://github.com/JuliaGraphics/Colors.jl/pull/482#issuecomment-864501967
```julia
julia> @btime convert.(RGB, $xyz_f32);
  36.715 ms (2 allocations: 11.44 MiB) # before (PR #482)
  31.418 ms (2 allocations: 11.44 MiB) # after

julia> @btime convert.(RGB{N0f8}, $xyz_f32);
  42.136 ms (2 allocations: 2.86 MiB) # before (PR #482)
  40.123 ms (2 allocations: 2.86 MiB) # after
```

kimikage:
The connection between the curve and the line in `fxyz2lab` exactly matches in value and slope. Therefore, it is possible to merge the two with `min`. On the other hand, in `srgb_compand` there is a small gap between the curve and the line.
```julia
julia> Float64(big"0.0031308"^(5/big"12") * big"1.055" - big"0.055")
0.04044990748269015

julia> Float64(big"0.0031308" * big"12.92") # != 0.04045
0.040449936
```

Issue #477, opened 2021-05-14
Optimizing the types used for intermediate calculations


kimikage:
Currently, many Float64-based calculations are used for color space conversions and color difference calculations, regardless of the input/output type. However, `Float32` is accurate enough for practical use in many color and image related applications.

I intend to speed up `RGB{N0f8}`-->`XYZ{Float32}`-->`Lab{Float32}` conversions and `colordiff`.

Probably the breaking change is that the `colordiff` for `RGB{N0f8}` will return `Float32` instead of `Float64`.

kimikage:
BTW, I was wondering why the required accuracy of `colordiff` (`DE_2000`) was so low.
https://github.com/JuliaGraphics/Colors.jl/blob/ddfc19bc3fb35500d4a2056900aec98763b13840/test/colordiff.jl#L48

And then I found two typos in the test data.
https://github.com/JuliaGraphics/Colors.jl/blob/ddfc19bc3fb35500d4a2056900aec98763b13840/test/colordiff.jl#L36-L37

((60.2574, -34.0099,  36.2"**6**"77), (60.4626, -34.1751,  39.4387),  1.2644),
((63.0109, -31.0961,  -5.8663), (62.8187, -29.7946,  -4.0864),  1.2"**6**"30),

The fixes will allow us to minimize the tolerance (i.e. `eps_cdiff = 0.00005`).

kimikage:
One of the most costly parts of the `DE_2000` is the following `t` ( *T* ) calculation.
https://github.com/JuliaGraphics/Colors.jl/blob/9f27da0271f8dcbeca5a0e100d7958f86a6e658d/src/differences.jl#L201-L206

When `t` is computed with `Float32`, the accuracy is about 8 ULP (the average is about 7 ULP), even though it requires about 40 ns.
I think this is a polynomial approximation of the whole `t`, rather than optimizing `cosd(::Float32)`, which would be more advantageous in terms of accuracy.
I don't have a good idea about the `Float64` case yet.

kimikage:
The accuracy bottleneck in calculating `DE_2000` is in `delta_h`. `delta_h` involves the calculation of `sin(Δh/2)`, where `Δh` is the angle between the ( a*, b*) vectors of the two colors.

[...]

PR #494, opened 2021-06-20
Use `Float32` to calculate `DE_2000` for low precision colors

kimikage:
This speeds up the calculation of `DE_2000` while ensuring the accuracy of `5e-5`.
This changes not only the type of intermediate variables, but also the return type. (cf. #477)
~This also fixes typos in the `DE_2000` test data.~ (separated to #499)

```julia
julia> a, b = Lab(50, -1.3802, -84.2814), Lab(50, 0.0000, -82.7485);

julia> colordiff(Lab{BigFloat}(a), Lab{BigFloat}(b)) |> Float64
0.9999988647524654

julia> colordiff(Lab{Float64}(a), Lab{Float64}(b))
0.9999988647524661

julia> colordiff(Lab{Float32}(a), Lab{Float32}(b))
1.0000001f0

julia> colordiff(Lab{BigFloat}.(Lab{Float32}.((a, b)))...) |> Float32
0.99999994f0
```

### Benchmark
```julia
julia> a = Lab{Float32}.(rand(RGB{Float32}, 1000, 1000));

julia> b = Lab{Float32}.(rand(RGB{Float32}, 1000, 1000));

julia> @btime colordiff.($a, $b);
  174.563 ms (2 allocations: 7.63 MiB) # v0.12.8 (`Float64`)
   69.342 ms (2 allocations: 3.81 MiB) # this PR (`Float32`)

julia> a64, b64 = Lab{Float64}.(a), Lab{Float64}.(b);

julia> @btime colordiff.($a64, $b64);
  107.792 ms (2 allocations: 7.63 MiB) # this PR

julia> versioninfo()
Julia Version 1.6.2
Commit 1b93d53fc4 (2021-07-14 15:36 UTC)
Platform Info:
  CPU: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, tigerlake)
```

### DIFF
--- a/docs/src/colordifferences.md
+++ b/docs/src/colordifferences.md
@@ -4,13 +4,13 @@ The [`colordiff`](@ref) function gives an approximate value for the difference b
 
 ```jldoctest example; setup = :(using Colors)
 julia> colordiff(colorant"red", colorant"darkred")
-23.75414245117878
+23.754143f0
 
 julia> colordiff(colorant"red", colorant"blue")
-52.881355694354724
+52.881363f0
 
 julia> colordiff(HSV(0, 0.75, 0.5), HSL(0, 0.75, 0.5))
-19.485908737785326
+19.48590873778533
 ```
 
 ```julia
--- a/src/differences.jl
+++ b/src/differences.jl
@@ -167,47 +167,49 @@ DE_DIN99o()
 
 
 # Delta E 2000
-function _colordiff(ai::Color, bi::Color, m::DE_2000)
-    twentyfive7 = pow7(25)
-
-    # Ensure that the input values are in L*a*b* space
-    a_Lab = convert(Lab, ai)
-    b_Lab = convert(Lab, bi)
+# Ensure that the input values are in L*a*b* space
+function _colordiff(a::Color, b::Color, m::DE_2000)
+    _colordiff(promote(convert(Lab, a), convert(Lab, b))..., m)
+end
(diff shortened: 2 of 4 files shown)


kimikage:
At the moment, this uses `Float32` entirely, in order to identify the effect of precompilation. However, in practice, using `Float32` instead of `Float64` outside of bottlenecks should not result in significant speed improvements.

Probably, `atand`s  and `cosd`s are the bottlenecks.

kimikage:
I would like to determine if it is worth calculating `delta_h` in `DE_2000` with `Float32`, after improving the accuracy of `delta_h`.
cf. https://github.com/JuliaGraphics/Colors.jl/issues/477#issuecomment-869102943

johnnychen94:
FWIW, I doubt that anyone is really using it for `DE_2000`, but CUDA devices are partialcularily optimized for Float32 operations and not for Float64.

kimikage:
Probably the last bottleneck is `mean_hue`. As we learned in high school, a vector bisecting the angle of two direction vectors can be represented by the composition of their unit vectors. This method reduces the costly `atan` to a single call, but causes a loss of significance when the mean hue is near 0°, 90°, 180°, or 270°. Perhaps rotating the entire field will work around the problem.

I've spent too much time improving `delta_h`, so I'd like to merge this PR first, with `mean_hue` being a future task.

kimikage:
Since this change is related to issue #496 as well as speed, I would like to merge this.

PR #497, opened 2021-06-27
Remove unused `import`

kimikage:
This is a matter of preference, but it's not a bad idea to make it clear which functions we are overloading as ”public" APIs.

No additional changes are planned at this time, but this should be merged in just before the release of v0.13.


kimikage:
Although this was overlooked in the v0.13.0 release, we need to prepare to switch to v0.14 to support FPN v0.9.

PR #506, opened 2021-07-31
Optimize `distinguishable_colors`

### LINKED ISSUE #477: Optimizing the types used for intermediate calculations

kimikage:
This changes the type used for internal operations from `Lab{Float64}` to `Lab{Float32}`.
This also removes the unnecessary sRGB gamma operations and `deleteat!`.

Note that `@nospecialize` is almost useless because of the keyword arguments.

```julia
julia> @btime distinguishable_colors(10);
  10.832 ms (896527 allocations: 14.88 MiB) # master w/o recompilation
  5.104 ms (50998 allocations: 1.50 MiB)    # master w/ recompilation
  2.675 ms (9070 allocations: 213.08 KiB)   # this PR
```
Closes #477

### DIFF
--- a/src/colormaps.jl
+++ b/src/colormaps.jl
@@ -4,6 +4,12 @@ include("maps_data.jl")
 # color scale generation
 # ----------------------
 
+@noinline function _generate_lab(l::Float32, c::Float32, h::Float32)
+    lab = convert(Lab{Float32}, LCHab{Float32}(l, c, h))
+    rgb = xyz_to_linear_rgb(convert(XYZ{Float32}, lab))
+    convert(Lab{Float32}, linear_rgb_to_xyz(correct_gamut(rgb)))
+end
+
 """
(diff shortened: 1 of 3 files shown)


kimikage:
BTW, the syntax for static parameters in the documentation is still outdated.
https://github.com/JuliaGraphics/Colors.jl/blame/f7fddbd3f84df60eb4a667428a900e121309f150/docs/src/colormapsandcolorscales.md#L220-L232

Issue #496, opened 2021-06-22
Fatal performance degradation due to precompilation or inference bugs


kimikage:
This is the follow-up to issue #425.

If precompilation is enabled (default), some functions will be boxed, causing significant performance degradation. For example:
```julia
julia> using Colors, BenchmarkTools # Colors.jl v0.12.8

julia> rgb_f64 = rand(RGB{Float64}, 1000, 1000);

julia> @btime convert.(XYZ, $rgb_f64);
  140.676 ms (5758522 allocations: 110.76 MiB)
```

This is a bug in Julia, although it has not been clearly identified, and several related issues have been reported.
- https://github.com/JuliaLang/julia/issues/34055
- https://github.com/JuliaLang/julia/issues/35537
- https://github.com/JuliaLang/julia/issues/35800
- https://github.com/JuliaLang/julia/issues/35972

"#35800" is often used as a synonym of this problem because of the memorability of the issue number.

I am not familiar with Julia's internals, so this looks very strange to me. However, I know a few things empirically.
Perhaps the root cause is that inferencing is context-dependent. A typical example of context-dependency is inlining. Inlining encourages constant propagation and makes inferences more likely to succeed.
We can't spend infinite time on the inference, so interrupting inferencing in a deep method chain is inevitable. The problem is that once a method fails to be inferred, it is marked as an "inference failure" until it is recompiled.

Perhaps Julia v1.6.x, the next LTS candidate, will not fully solve this problem. So we will have to live with this bug.

Of course, `__precompile__(false)` is a valid workaround. However, it is a last resort. There are two major approaches to workarounds. One is to make it easier for the inference to succeed (regardless of the context). The other is to avoid using "marked" (and likely to be "marked") functions.

I have implemented several workarounds in Colors v0.13 series, but the problems still remain.

**Edit:**
A new package JETTest.jl may have less side effects than the runtime-dispatch detection based on benchmarks (such as `@btime`).
https://discourse.julialang.org/t/ann-jettest-jl-advanced-testing-toolset-for-julia/63229

kimikage:
### Conversions

#### v0.12.8
| from\to | `RGB` | `HSV` | `XYZ` | `Lab` | `LCHab` | `Luv` | `LCHuv` | 
|:--------|:-----:|:-----:|:-----:|:-----:|:-------:|:-----:|:-------:| 
[...]

PR #508, opened 2021-08-02
Mitigate inference problems in color conversions

kimikage:
This uses the `cnvt` instead of the top-level `convert` when source and destination are both known colors.
This also blocks the propagation of inference failures by adding type assertions to the result of `convert`.
~However, these changes do not solve the essential problem (#496). So, this PR is for reference only, and I have no plans to merge this for now.~
By PR #510, we can see the practical effects, e.g.：
```julia
julia> lchab = rand(LCHab, 1000, 1000);

julia> convert.(RGB{Float64}, lchab); # compilation

julia> @time convert.(RGB{Float64}, lchab);
  0.174340 seconds (4.00 M allocations: 144.959 MiB, 37.16% gc time) # before
  0.137165 seconds (2.00 M allocations: 83.923 MiB, 40.44% gc time)  # after
```
I haven't solved the problem yet, though.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -59,6 +59,10 @@ convert(::Type{XYZ{T}}, c, wp::XYZ) where {T} = cnvt(XYZ{T}, c, wp)
 convert(::Type{Lab{T}}, c, wp::XYZ) where {T} = cnvt(Lab{T}, c, wp)
 convert(::Type{Luv{T}}, c, wp::XYZ) where {T} = cnvt(Luv{T}, c, wp)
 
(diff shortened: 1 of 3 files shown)


kimikage:
Some problematic math functions have been removed from the major paths. The problems that still remain are trigonometric functions, especially `sincos`.
Since `sincos` is not that bad in terms of speed and accuracy, there is not enough motivation to reimplement it.

kimikage:
`RGB` <--> `LCHab`/`LCHuv` conversions still have problems. However, I would like to address some of the problems, in PR #509.

Issue #300, opened 2017-11-04
does mid work on colormap?


ultradian:
I was playing with trying to change the midpoint in a colormap and it doesn't work the way I expected.

For example, `colormap("Blues")` generates an array that starts with 
```
 RGB{Float64}(0.958129,0.992733,0.998897) 
 RGB{Float64}(0.948956,0.98958,0.997057)  
 RGB{Float64}(0.939677,0.986361,0.995234) 
```
and goes to
```
 RGB{Float64}(0.0548094,0.182609,0.382798)
 RGB{Float64}(0.0484511,0.169616,0.362331)
 RGB{Float64}(0.0418427,0.156645,0.341597)
```
with midpoint `colormap("Blues")[50]` giving:
`
RGB{Float64}(0.40043042841321896,0.7304593324507547,0.9058502694643216)
`
When I try `colormap("Blues", mid=0.75)[50]` I get the same result.  I thought I should get something different because this point would move to the left.  Do I misunderstand something?

timholy:
Hoping someone who uses colormaps replies here. CC @glenn-sweeney

PR #509, opened 2021-08-02
Optimize `colormap`

kimikage:
This uses `Tuple` instead of `Vector` for the `colormap` parameters. (cf. https://github.com/JuliaGraphics/Colors.jl/commit/ddfe8d5f3aa69bdc566d28c82b81446575735a86)
This also unifies the type of preset values to `Float64`.

However, at present, there is no speed improvement. :sob:
I will make some more tweaks and add some tests.
This is more of a compilation time issue than a runtime performance issue. I am guessing that there is some relation between the precompilation problem with `colormap` and #496.
#### before
```julia
julia> using Colors

julia> @time colormap("RdBu", 100);
  0.356729 seconds (1.47 M allocations: 87.943 MiB, 8.93% gc time, 99.93% compilation time)

julia> @time colormap("RdBu", 100);
  0.000067 seconds (916 allocations: 39.406 KiB)

julia> @time colormap("Blues", 100);
  0.000056 seconds (656 allocations: 25.875 KiB)

julia> @time colormap("RdBu", 100, logscale=true);
  0.009631 seconds (55.90 k allocations: 3.555 MiB, 99.07% compilation time)

julia> @time colormap("RdBu", 100, mid=0.7);
  0.000111 seconds (916 allocations: 41.219 KiB)
```
#### after (w/o PR #511)
```julia
julia> using Colors

julia> @time colormap("RdBu", 100);
  0.069386 seconds (605.76 k allocations: 38.593 MiB, 17.43% gc time, 99.91% compilation time)

julia> @time colormap("RdBu", 100);
  0.000025 seconds (30 allocations: 6.031 KiB)

julia> @time colormap("Blues", 100);
  0.000021 seconds (16 allocations: 3.000 KiB)

julia> @time colormap("RdBu", 100, logscale=true);
  0.012442 seconds (55.02 k allocations: 3.522 MiB, 99.67% compilation time)

julia> @time colormap("RdBu", 100, mid=0.7);
  0.000029 seconds (30 allocations: 6.016 KiB)
```
#### after (w/ PR #511)
```julia
julia> using Colors

julia> @time colormap("RdBu", 100);
  0.000042 seconds (6 allocations: 5.281 KiB)

julia> @time colormap("Blues", 100);
  0.000019 seconds (4 allocations: 2.625 KiB)

julia> @time colormap("RdBu", 100, logscale=true);
  0.008044 seconds (54.99 k allocations: 3.521 MiB, 99.62% compilation time)

julia> @time colormap("RdBu", 100, mid=0.7);
  0.000032 seconds (6 allocations: 5.266 KiB)
```

This also fixes the problem with hue interpolation.
colormap_hue


src/colormaps.jl
```diff
@@ -280,49 +284,33 @@ Extra control is provided by keyword arguments.
...
-        p = copy(colormaps_diverging[cname][1:11])
-
-        for (k,v) in kvs
-            ind = findfirst(e->e==k, allowedkeys)
-            ind == nothing && throw(ArgumentError("Unknown keyword argument $k"))
-            if ind > 0
-                p[ind] = v
-            end
+        ps(i) = oftype(pbs[i], get(kvs, keys_s[i], pbs[i]))
+        return _sequential_palette(N, logscale, (ps(i) for i in eachindex(pbs))...)
```
johnnychen94, line 301: What's the advantage of this generator compared to the following?

```julia
        return _sequential_palette(N, logscale, map(ps, eachindex(pbs))...)
```

src/colormaps.jl
```diff
@@ -280,49 +284,33 @@ Extra control is provided by keyword arguments.
...
+function colormap(cname::AbstractString, N::Integer=100; kvs...)
+    _colormap(String(cname), Int(N), kvs)
+end
+function _colormap(cname::String, N::Int, kvs)
+    logscale = get(kvs, :logscale, false)::Bool
+    lcname = lowercase(cname)
+    if haskey(colormaps_sequential, lcname)
+        pbs = colormaps_sequential[lcname]
+        keys_s = (:h, :w, :d, :c, :s, :b, :wcolor, :dcolor)
+        for k in keys(kvs)
```
johnnychen94, line 296: This is likely not a performance hotspot. Will this check be redundant? Maybe something like the following?

```julia
k in intersect(keys_d, keys(kvs)) || k in (:logscale, )
```

src/colormaps.jl
```diff
@@ -116,81 +116,85 @@ that are then blended to the original hue smoothly.
...
+        ll = bezier(tt, p0.l, p2.l, q0.l, q1.l, q2.l)
+        cc = bezier(tt, p0.c, p2.c, q0.c, q1.c, q2.c)
+        hh = bezier(tt, p0.h, p2.h, q0.h, q1.h, q2.h)
 
-        push!(pal, convert(RGB, LCHuv(ll,cc,hh)))
+        @inbounds pal[i] =LCHuv{Float64}(ll, cc, hh)
     end
 
-    pal
+    RGB{Float64}.(pal)
```
johnnychen94, line 197: This is suspicious to me. Any reason why we can't do allocating `pal` as `RGB{Float64}` vector, and do the conversion in the for loop?

src/colormaps.jl
```diff
@@ -200,53 +204,53 @@ Create diverging palettes by combining 2 sequential palettes
...
 
-    pal2 = sequential_palette(h2, N2+1, w=w, d=d2, c=c, s=s, b=b, wcolor=wcolor, dcolor=dcolor2, logscale=logscale)
+    pal1 = _sequential_palette(N1 + 1, logscale, h1, w, d1, c, s, b, wcolor, dcolor1)
+    pal2 = _sequential_palette(N2 + 1, logscale, h2, w, d2, c, s, b, wcolor, dcolor2)
 
     if isodd(N)
-        midcol = weighted_color_mean(0.5, pal1[end], pal2[1])
-        return [pal1[1:end-1]; midcol; pal2[2:end]]
+        midcol = weighted_color_mean(0.5, pal1[1], pal2[1])
+        return [pal1[end:-1:2]; midcol; pal2[2:end]]
```
johnnychen94, line 251: `pal2[2:end]` is unnecessary memory allocation, we can use `view(pal2, 2:end)` instead.

src/colormaps.jl
kimikage: It "looks" good. However, doesn't `map` return a `Vector{Union{Float64, RGB{Float64}}}` in this case? I haven't checked the native code enough, but I think it puts a load on the optimizer.

Of course, I think it would have little effect on "runtime" costs.

johnnychen94: I thought `pbs` is a `Tuple` here?

src/colormaps.jl
kimikage: Personally, I find the procedural approach easier to understand.

Note that it is possible to kick out `logscale` from key-value pairs.
```julia
colormap(cname::AbstractString, N::Integer=100; logscale=false, kvs...)
```
I had hoped that avoiding this separation would reduce the load of inference and compilation, but it seems to have no effect. (Again, `@nospecialize` doesn't help.)
However, I thought I'd address the problem of not getting an error when `mid` is specified for a sequential colormap.  (#300)

src/colormaps.jl
kimikage: It's hard to say which is more efficient. Eventually we should do some benchmarking to make sure, but The `LCHuv`->`RGB` conversion is still not "safe" enough in terms of #496.

src/colormaps.jl
kimikage: That's right, but `eachindex(::Tuple)` is a `Base.OneTo`.

src/colormaps.jl
johnnychen94: Hmmm, the memory allocation that `RGB{Float64}.(pal)` introduces seems unnecessary to me.

kimikage: Yes, it is definitely unnecessary. :sweat_smile:
Keeping the conversion out of the loop makes it easier to isolate the problems.

src/colormaps.jl
johnnychen94: `ntuple(ps, length(pbs))` maybe? Since it's a tuple, we can assume it to be 1-based. And I believe Julia can optimize Tuple better than Generator.

Again, I didn't do profile here so it's perhaps not the hotspot. I just point out things that I think have room for a better version.

src/colormaps.jl
kimikage: In fact, it does not make a big difference in time for an array length of a few hundred. However, `@views` doesn't harm readability much, so I added it.

src/colormaps.jl
kimikage: I think `ntuple` is for homogeneous tuples. I used `Tuple` to add type annotations. However, the generator itself doesn't seem to be too bad.

kimikage: BTW, in the sequential palette case, the optimization may be working well because the number of parameters (i.e. tuple elements) happens to be `2^3`.

src/colormaps.jl
kimikage: PR #510 may be helpful in changing the `LCHuv` --> `RGB` conversion back to the inner loop.

kimikage: It's back (with an ugly workaround).:smile:

kimikage:
Remaining problems:
- Precompilation of `RGB`-->`LCH` conversion (PR #511 can fix this)
- Over-specialization with keyword arguments
- Precompilation of `vcat` in `_diverging_palette`

These can be tackled in separate PRs, though.

src/colormaps.jl
kimikage: Since `vcat` and `copyto!` cannot be precompiled, I ended up using a `for` loop. :sweat_smile:

PR #495, opened 2021-06-22
Optimize `ColorTypes.hue(c)`

kimikage:
The conventional version using `atand` has a larger numerical error because `rad2deg` is applied to the `atan` result.
This improves the accuracy and speed by making the internal representation conform to degrees and by simplifying the handling of exceptional cases.

This change should be made in ColorTypes.jl, but for the convenience of development, it is implemented once in Colors.jl.

```julia
julia> hue(Lab(50, big(50.85106f0), big(-24.293373f0))) |> Float32
334.4645f0

julia> hue(Lab(50, 50.85106f0, -24.293373f0))
334.4645f0

julia> hue(ALab(50, 50.85106f0, -24.293373f0)) # all versions except `Lab` and `Luv` are old at the moment
334.46454f0
```

PR #510, opened 2021-08-09
Implement `polar_to_cartesian`

kimikage:
This adds `sincos360`, the rough equivalent of `sincosd`.
The `sincos360` is less accurate than `sincosd`, but in many cases more accurate than `sincos(deg2rad(x))`.
Also, while `sincosd(Inf)` throws a `DomainError`, `sincos360(Inf)` returns `(NaN, NaN)`.

This is mainly intended to mitigate precompilation and inlining issues. (cf. issue #496)

This might have to be moved to ColorTypes.jl as well as `atan360`. (cf. PR #495) Fortunately, moving those methods to ColorTypes.jl seems to maintain the benefits. However, I'm not sure why the problem occurs with `Base`.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -200,8 +200,7 @@ function MSC(h)
     end
     p, t = pt(h)
 
-    alpha=-sind(h)
-    beta=cosd(h)
+    beta, alpha = polar_to_cartesian(1.0, -h)
 
     # un & vn are calculated based on reference white (D65)
     un, vn = xyz_to_uv(WP_DEFAULT)
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -468,8 +468,7 @@ end
 
 
 function cnvt(::Type{Lab{T}}, c::LCHab) where T
-    b, a = c.c .* sincos(deg2rad(c.h))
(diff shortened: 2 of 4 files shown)


kimikage:
```julia
julia> lchab32 = [LCHab(100*rand(Float32), 100*rand(Float32), 360*rand(Float32)) for y = 1:1000, x = 1:1000];

julia> lchab64 = [LCHab(100*rand(Float64), 100*rand(Float64), 360*rand(Float64)) for y = 1:1000, x = 1:1000];

julia> @btime convert.(Lab{Float32}, $lchab32);
  30.570 ms (2000002 allocations: 41.96 MiB) # before
  12.473 ms (2 allocations: 11.44 MiB) # before (recompiled)
  8.589 ms (2 allocations: 11.44 MiB)  # after

julia> @btime convert.(Lab{Float64}, $lchab64);
  30.995 ms (2000002 allocations: 68.66 MiB) # before
  16.649 ms (2 allocations: 22.89 MiB) # before (recompiled)
  10.462 ms (2 allocations: 22.89 MiB) # after
```

kimikage:
This has almost no impact on functionality, so I will merge this in order to proceed with other tasks.

PR #511, opened 2021-08-27
Work around #496 for Luv, Lab

timholy:
This fixes the specific demo in #496. These constants were added in 8f6550dd, but there is nothing wrong with that PR---those should be perfectly reasonable operations to perform during package definition.

The origin of the problem is a julia bug whose identity remains mysterious. Some diagnostics were done in https://github.com/JuliaLang/julia/issues/35972, but this particular case appears to be distinct from the mechanism identified there and has not yet been isolated.  This is a simple workaround.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -151,12 +151,12 @@ tritanopic(c::Color)   = tritanopic(c, 1.0)
 # MSC - Most Saturated Colorant for given hue h
 # ---------------------
 
-const LUV_HUE_R = hue(convert(Luv, RGB(1.0, 0.0, 0.0)))
-const LUV_HUE_Y = hue(convert(Luv, RGB(1.0, 1.0, 0.0)))
-const LUV_HUE_G = hue(convert(Luv, RGB(0.0, 1.0, 0.0)))
-const LUV_HUE_C = hue(convert(Luv, RGB(0.0, 1.0, 1.0)))
-const LUV_HUE_B = hue(convert(Luv, RGB(0.0, 0.0, 1.0)))
-const LUV_HUE_M = hue(convert(Luv, RGB(1.0, 0.0, 1.0)))
+const LUV_HUE_R = 12.17397852379156 # hue(convert(Luv, RGB(1.0, 0.0, 0.0)))
+const LUV_HUE_Y = 85.87273351614108 # hue(convert(Luv, RGB(1.0, 1.0, 0.0)))
+const LUV_HUE_G = 127.72355232980077 # hue(convert(Luv, RGB(0.0, 1.0, 0.0)))
+const LUV_HUE_C = 192.17397852379156 # hue(convert(Luv, RGB(0.0, 1.0, 1.0)))
+const LUV_HUE_B = 265.8727335161411 # hue(convert(Luv, RGB(0.0, 0.0, 1.0)))
+const LUV_HUE_M = 307.7235523298008 # hue(convert(Luv, RGB(1.0, 0.0, 1.0)))
 """
     MSC(h)
     MSC(h, l; linear=false)
@@ -259,7 +259,7 @@ function find_maximum_chroma(c::C,
     end
 end
 
-const LAB_HUE_Y = hue(convert(Lab, RGB(1.0, 1.0, 0.0)))
+const LAB_HUE_Y = 102.85123437653252 # hue(convert(Lab, RGB(1.0, 1.0, 0.0)))
 
 function find_maximum_chroma(c::LCHab{T}) where T
     maxc = find_maximum_chroma(c, 0, 135)


kimikage:
I am interested in how you found this workaround. Also, I would like to know the specific cases where this workaround is effective.
I might be able to suggest another workaround.

timholy:
Ah, very simple actually. `precompile(f, argtyps)` simply forces `f` to be compiled for `f(::argtyps...)`, nothing more. So does executing `f(args::argtyps...)`. Consequently, after commenting out the explicit `precompile` directives in Colors, ColorTypes, and FixedPointNumbers failed to change anything, I went looking for other examples of statements that executed code while the package was being defined (i.e., before it gets cached to disk). Method definitions, of course, don't actually run any code unless they are being defined by metaprogramming.

timholy:
Following up: `precompile` has nothing to do with caching to the `*.ji`. It just compiles code. This is almost the same as running code, except that `precompile` only works until it gets to a runtime dispatch barrier and then quits, whereas running code executes the runtime dispatch and then continues to compile as needed for the next steps of execution. Hence running code is often a more thorough version of `precompile`. This is part of why in my more recent precompilation work I've tended to just supply tiny workloads (e.g., https://github.com/JuliaImages/ImageCore.jl/blob/ba4abdd6c9d30ffd9360a4b0f9e22636bfb3cf2f/src/precompile.jl#L51-L108).

The caching is an independent step that caches all methods + their specializations defined in the package. Anything that gets compiled is a specialization, so `precompile == running code` for the purposes of caching.

But actually, the paragraph above this one makes me think: given all the inference failures in https://github.com/JuliaGraphics/ColorTypes.jl/issues/266, would some of our weird issues be fixed by switching from `precompile(Tuple{typeof(convert),Type{C{T}},XYZ{T}})` to something like `convert(C{T}, zero(XYZ{T}))` in https://github.com/JuliaGraphics/Colors.jl/blob/master/src/precompile.jl? It won't cache things in ColorTypes, unfortunately (not without https://github.com/JuliaLang/julia/pull/42016). Worth testing, but I'm already running behind on some other deadlines and have to move on for the next couple of days.

kimikage:
Thank you for the detailed explanation. It is very logical. However, I feel that the effect of this PR is somewhat less than the logical expectation.

Of course, since the rewriting of the core code of conversions is mostly complete, there is no major disadvantage to making the `const`s literals.

timholy:
I'm in no hurry to merge this, if you want to investigate more. Just wanted to make sure you knew about the causes & potential workarounds.

kimikage:
If we use `BigFloat` to calculate the constants, the execution may not harm the code for `Float32`/`Float64`. It's ugly, though.

kimikage:
> switching from `precompile(Tuple{typeof(convert),Type{C{T}},XYZ{T}})` to something like `convert(C{T}, zero(XYZ{T}))`

It seems that a simple switching does not work well. :thinking: It might be worth looking into the differences with `precompile` a bit more.

timholy:
>  It might be worth looking into the differences with precompile a bit more

The reason for the difference is known (`precompile` only works up until runtime-dispatch "barriers," whereas execution continues through them), but the consequences of the difference may be revealing. Given that we now know that `convert` has so many *internal* inference failures (https://github.com/JuliaGraphics/ColorTypes.jl/issues/266, even though the final result is inferrable), this is interesting but not surprising. But it might indeed help further diagnose Julia's own issues here.

kimikage:
For example, if we don't call `cnvt(Luv{Float64}, ::XYZ{Float64})`, the problem seems to be avoided. However, while that is superficially obvious, it is still essentially mysterious.
```julia
function _get_hue_in(::Type{C}, r, g, b) where C <: Union{Luv, Lab}
    xyz = linear_rgb_to_xyz(RGB{Float64}(r, g, b))
    wp = WP_DEFAULT
    if C === Luv
        c = xyz2luv(Luv{Float64}, xyz, xyz.y, xyz_to_uv(wp))
    else
        c = xyz2lab(Lab{Float64}, XYZ(xyz.x / wp.x, xyz.y, xyz.z / wp.z))
    end
    hue(c)
end

const LUV_HUE_R = _get_hue_in(Luv, 1, 0, 0)
...
```
Unlike `Base.convert`, `ColorTypes.cconvert`, etc., `cnvt` and `xyz2luv` are internal methods of `Colors`.
Also, `broadcast` (`.` operations) seems to have an indirect effect, but I haven't isolated the factors yet.

kimikage:
I see some room for improvement in the implementation of `Luv` conversions, but the compilation-related mystery still deepens.

I will merge #509 after merging this.

timholy:
> compilation-related mystery still deepens

I'm not sure it's much of a mystery anymore, it's just a difficult fix.

PR #513, opened 2021-08-28
Fix an ambiguity with AbstractTrees

### LINKED ISSUE #512: Ambiguity with AbstractTrees
[AbstractTrees]() is a dependency of SnoopCompile. Unfortunately, loading it and Colors at the same time introduces an ambguity that stands in the way of running this test suite:

```julia
julia> tinf = @snoopi_deep include("runtests.jl")
Test Failed at /home/tim/.julia/dev/Colors/test/runtests.jl:2
  Expression: isempty(detect_ambiguities(Colors, Base, Core))
   Evaluated: isempty(Tuple{Method, Method}[(getindex(M::Colors.Mat3x3{T}, i) where T in Colors at /home/tim/.julia/dev/Colors/src/utilities.jl:10, getindex(x::AbstractArray, ::AbstractTrees.ImplicitRootState) in AbstractTrees at /home/tim/.julia/packages/AbstractTrees/oHb1F/src/iteration.jl:103)])
ERROR: LoadError: There was an error during testing
```

timholy:
The fix is to adhere more closely to the minimalistic
[AbstractArray interface](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array).

Fixes #512

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,5 @@
 using Colors, Test
-@test isempty(detect_ambiguities(Colors, Base, Core))
+@test isempty(detect_ambiguities(Colors))
 
 include("algorithms.jl")
 include("conversion.jl")
@@ -10,3 +10,6 @@ include("din99.jl")
 include("display.jl")
 include("parse.jl")
 include("utilities.jl")
+
+using AbstractTrees
+isempty(detect_ambiguities(Colors))
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -12,4 +12,4 @@ include("parse.jl")
 include("utilities.jl")
 
 using AbstractTrees
-isempty(detect_ambiguities(Colors))
+@test isempty(detect_ambiguities(Colors))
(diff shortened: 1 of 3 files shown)


kimikage:
`AbstractTrees` is a widely used package, so I have no objection to this PR itself. However, I wonder about the scope of the package that should be tested.

test/runtests.jl
```diff
@@ -10,3 +10,6 @@ include("din99.jl")
 include("display.jl")
 include("parse.jl")
 include("utilities.jl")
+
+using AbstractTrees
+isempty(detect_ambiguities(Colors))
```
kimikage, line 15: Is it intentional that this exists independently of the ambiguity test above?

timholy: I decided to separate it simply because it's a bit different in nature. As you notice in https://github.com/JuliaGraphics/Colors.jl/pull/513#issuecomment-907587280, there's no limit to the number of packages we *could* test against. The early `detect_ambiguities` is kind of the minimal version ("must have"); the later one is potentially the "maximal" version ("nice to have").

But haha, I notice I didn't put this inside `@test`. Will fix.

kimikage (approved):

PR #515, opened 2021-09-06
Update documentation for colormaps

### LINKED ISSUE #265: Example sequential_palette
Hello,

I'm having some troubles generating a sequential_palette for a heat map. The documentation is not clear to me. Can someone provide me with a working example of sequential_palette and diverging_palette? Can I use this with a heatmap in Plots.jl? 

```julia
sequential_palette(h, [N::Int=100; c=0.88, s=0.6, b=0.75, w=0.15, d=0.0, wcolor=RGB(1,1,0), 
dcolor=RGB(0,0,1), logscale=false])

diverging_palette(h1, h2 [, N::Int=100; mid=0.5,c=0.88, s=0.6, b=0.75, w=0.15, d1=0.0, d2=0.0,
 wcolor=RGB(1,1,0), dcolor1=RGB(1,0,0), dcolor2=RGB(0,0,1), logscale=false])
```

kimikage:
This adds visual examples for `sequential_palette`.

sequential_palette

Closes #265

### DIFF
--- a/docs/colormapparams.jl
+++ b/docs/colormapparams.jl
@@ -1,0 +1,183 @@
+module ColormapParams
+
+using Colors
+using Main.PNG16x16
+
+struct ColormapParamSVG <: Main.SVG
+    buf::IOBuffer
+end
+
+function ColormapParamSVG(target::Symbol)
+    io = IOBuffer()
+    write(io,
+        """
+        <svg xmlns="http://www.w3.org/2000/svg"
+             xmlns:xlink="http://www.w3.org/1999/xlink"
+             version="1.1"
+             width="45mm" height="45mm"
+             viewBox="0 0 270 270" preserveAspectRatio="none"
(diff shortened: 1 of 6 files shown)


kimikage:
I had originally planned to update the documentation after #509. However, now I think it's better to update the documentation first to check the effect of the hue interpolation fix in #509.

kimikage:
Probably, ColorSchemes.jl is the most comprehensive and widely used package in visualization. As a matter of fact, we no longer have situations where we must use the `colormap`. Also, modern perceptual color maps, like "Viridis", no longer use the `Luv` color space. (Even the CAM02-UCS is already becoming outdated.)

So, I think we should regard `colormap` as a "simple" tool which can be used without loading the plot library.
I'm also starting to wonder if it might be a good idea to add predefined diverging colormaps, mimicking ColorBrewer.

PR #518, opened 2022-01-01
Add conversions to and from Oklab and LCHOklab

nealmckee:
Currently failing as the type doesn’t exist yet.

To and from XYZ, everything else through XYZ.

I’m using the original specifications provided here: https://bottosson.github.io/posts/oklab/

This is my first contribution to Julia, or any software package for that matter, so any help would be greatly appreciated. For example, should I also add this to the docs?

See also https://github.com/JuliaGraphics/ColorTypes.jl/pull/271


Artoria2e5:
LCHOklab is known as Oklch now after CSS4.

timholy:
Yes. Not quite sure I understand, are you volunteering @jariji ?

timholy:
This needs tests.

kimikage:
PR #533 was merged.

PR #522, opened 2022-08-30
add note on transparent - opaque conversions

t-bltg:
`transparent` and `opaque` should be appear in doc search results.


kimikage:
Ugh, we can't use `color()` in the context.

PR #521, opened 2022-05-21
Update a few cases where array indices are assumed

brenhinkeller:
This may not be everything, but a few simple cases that were easy to update

PR #527, opened 2022-12-10
fix union{} ambiguity

Roger-luo:
See https://github.com/JuliaLang/julia/issues/47840

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ docs/build/
 docs/site/
 .DS_Store
 /docs/Manifest.toml
+Manifest.toml
--- a/src/display.jl
+++ b/src/display.jl
@@ -33,13 +33,13 @@ end
 
 
 function Base.show(io::IO, mime::MIME"image/svg+xml", cs::AbstractVector{T};
-                   max_swatches::Integer=0) where T <: Color
+                   max_swatches::Integer=0) where {Union{} <: T <: Color}
     mat = Base.ReshapedArray(cs, (1, length(cs)), ())
     show(io, mime, mat, max_swatches=max_swatches)
 end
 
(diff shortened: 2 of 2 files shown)


Roger-luo:
I'm wondering if someone could review this? The test coverage drop does not seem related to this PR.

johnnychen94:
will make a release later

Roger-luo:
@aplavin yeah I think you are right, I just realize `Union{} <: T` is always true, thus this won't work unless we type pirate `Union{}` this MIME type. 

But this does not work for `show` since lots of backends will call the corresponding `show` by checking if there is a valid `show` overloaded for the richest MIME type, e.g https://github.com/fonsp/Pluto.jl/blob/0f0b6f596fd4cf27b2665e1d93ddd44d96fe2a99/src/runner/PlutoRunner.jl#L1148 as a result we will see `MIME"text/plain"` displayed as picture if we overload

```
MIME"image/svg+xml" => MIME"text/plain"
```

I don't have a good way to fix this now.

johnnychen94:
@aplavin Thanks for identifying this. I didn't think carefully when merging it -- was checking it on the phone and did a quick merge. Reverted in 9518578

This `Union{}` would be a nightmare to JuliaImages. Almost every function is written in such a way.

johnnychen94:
I realize that I shouldn't make the release on the master branch since there are plenty of breaking changes made by @kimikage since v0.12.8. Thus I've made a backport release branch https://github.com/JuliaGraphics/Colors.jl/tree/release-0.12 with a new patch release v0.12.10 to supersede v0.12.9.

v0.12.10 includes #521 (@brenhinkeller) and #525(@williamjsdavis) as bug fixes.

I've also submitted a yank patch to General https://github.com/JuliaRegistries/General/pull/74192

PR #529, opened 2023-01-04
Switch to SnoopPrecompile

timholy:
On my machine, the workload inside the `@precompile_all_calls`
goes from 0.3s with the previous precompile script to 0.08s with the new
one, a 4x reduction. Given that we don't have precompiles anymore in
ColorTypes (see https://github.com/JuliaGraphics/ColorTypes.jl/issues/270,
a decision I agree with), this package seems like a good place to have
some.


timholy:
On `nightly`, I can't detect any difference in load time between `master` and this branch: master is 0.189s, this branch is 0.186s.

PR #540, opened 2024-05-01
Update workflows

kimikage:

PR #543, opened 2024-05-02
Update workflows (Backport #540 and #541)

kimikage:
These changes are mostly unnecessary, but it is worth backporting just for `workflow_dispatch`.


kimikage:
The tests, the tests are broken...

PR #553, opened 2024-05-14
Use WebP for CVD simulation demo in docs instead of indexed PNGs

kimikage:
cf. https://github.com/JuliaGraphics/Colors.jl/pull/413#issuecomment-598112465
When I added this demo, I needed to use ImageMagick.jl to read and write compressed images. IIRC, there were no JLLs at that time.


kimikage:
cf. https://discourse.julialang.org/t/what-are-the-best-practices-for-using-incompatible-downstream-packages-in-package-development/114275

ViralBShah:
Good to merge?

PR #557, opened 2024-10-20
Remove deprecated bindings RGB4 RGB1

dgleich:
This removes the deprecated bindings for RGB4 and RGB1, see discussion:

https://github.com/JuliaIO/ImageMagick.jl/issues/235
and similar fix in ColorTypes.jl
https://github.com/JuliaGraphics/ColorTypes.jl/pull/312

It also requires the new version of ColorTypes in Compat, but that will cause failures until that version is registered. 

CC: @ViralBShah, @ararslan

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,7 +8,7 @@ FixedPointNumbers = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
-ColorTypes = "0.11.5"
+ColorTypes = "0.12"
 FixedPointNumbers = "0.6, 0.7, 0.8"
 Reexport = "0.2, 1.0"
 julia = "1"
--- a/src/Colors.jl
+++ b/src/Colors.jl
@@ -4,8 +4,10 @@ using FixedPointNumbers
 using Reexport
 
 @reexport using ColorTypes
-Base.@deprecate_binding RGB1 XRGB
-Base.@deprecate_binding RGB4 RGBX
+# remove deprecated bindings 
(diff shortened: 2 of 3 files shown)


ViralBShah:
ColorTypes.jl registration PR: https://github.com/JuliaRegistries/General/pull/117694

Let's re-run the CI in an hour or so.

ViralBShah:
There is an ambiguity test failing:
```
     Testing Running tests...
Test Failed at /home/runner/work/Colors.jl/Colors.jl/test/runtests.jl:2
  Expression: isempty(detect_ambiguities(Colors))
   Evaluated: isempty(Tuple{Method, Method}[(_convert(::Type{Cout}, ::Type{C1}, ::Type{C1}, c) where {Cout<:Color, C1<:Color} @ ColorTypes ~/.julia/packages/ColorTypes/0RO5O/src/conversions.jl:90, _convert(::Type{Cdest}, ::Type{Odest}, ::Type{Osrc}, g) where {Cdest<:Color, Odest, Osrc<:AbstractGray} @ Colors ~/work/Colors.jl/Colors.jl/src/conversions.jl:75)])
```

ViralBShah:
Commenting out the ambiguity test, everything else runs fine.

timholy:
Let me take a crack at fixing the ambiguity.

timholy:
Personally I think we should remove the Invalidations workflow, but we can do that later.

timholy:
Does this look good to you @ViralBShah? I haven't looked at other PRs. The docbuild failure is unfortunate but is perhaps better fixed separately.

ViralBShah:
The doc changes are formatting related, and should be straightforward to fix separately.

PR #560, opened 2024-10-22
Update doc formatting to Julia 1.10

timholy:
Julia has periodically changed its print formatting. This updates to
Julia 1.10, although the doctests also pass on Julia 1.11. doctests
won't pass on old versions of Julia, but since 1.10 is the new LTS it
makes sense for this to be the default.

Also:
- explicitly use `lts` in the documentation workflow
- add `[compat]` in `docs/Project.toml`
- adjust for differences in CPU rounding in doctests
- fix an outdated reference to Juno


ViralBShah:
Could you remove invalidations from the CI in this PR too?

PR #565, opened 2025-02-27
remove invalid fastmath usage in atan360

KristofferC:
The functions need to be able to deal with infinities but fastmath assumes that arguments and results are not so.

On 1.12:

```
julia> Colors.atan360(-Inf, Inf)
NaN
```

with fastmath removed:

```
julia> Colors.atan360(-Inf, Inf)
315.0
```

I haven't looked into any of the other fastmath usages in this package.


ViralBShah:
Merge?

johnnychen94 (approved):

PR #566, opened 2025-02-27
avoid stackoverflow when passing two types that can not be promoted to a matching types in `mean_hue`

KristofferC:
The test at https://github.com/JuliaGraphics/Colors.jl/blob/f143ad9e9e457d0b1c8f94cca60bbc67746402f7/test/utilities.jl#L223 runs a function that stack overflows. Continuing the process after a stack overflow is not recommended and julia 1.12 prints:

```
Warning: detected a stack overflow; program state may be corrupted, so further execution might be unreliable.
```

Instead, catch if we have no match after the promote and throw a normal error.

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -395,14 +395,17 @@ When the hue difference is exactly 180 degrees, which of the two mean hues is
 returned depends on the implementation. In other words, it may vary by the color
 type and version.
 """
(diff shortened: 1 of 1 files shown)


ViralBShah:
Merge?

johnnychen94 (approved):

PR #571, opened 2025-07-03
fixes broken example for "weighted mean of more than two colors"

alex-s-gardner:
old example no longer works


kimikage:
When was it deprecated in the first place?
Since I'm now away from Colors.jl, I don't know where that discussion took place.

alex-s-gardner:
I'm not sure deprecated is the right description. Changing title to fixes "broken" example

kimikage:
How did you determine that it was broken?
The document generation passes even without this PR.

And if there is no consensus on the deprecation, we should consider whether we can fix the implementation before updating the documentation.

PR #579, opened 2026-09-20
Update workflows

kimikage:


kimikage:
The documentation error will be fixed in PR #580.
The `DE_BFD` error appears to be a separate issue.

